PR #5, opened 2018-06-15
PHPCS and comments

jameswburke:
* Recursive function for building menus.
* Hooks to modify menu shapes.


inc/components/class-menu.php
```diff
@@ -37,6 +37,18 @@ public function default_config() {
...
+
+		// Get menu locations.
+		$locations = get_nav_menu_locations();
+
+		// Get object id by location.
+		$menu_term = wp_get_nav_menu_object( $locations[ $menu_location ] ?? null );
+
+		// If valid menu term.
+		if ( $menu_term instanceof \WP_Term ) {
+			$this->parse_wp_menu( $menu_term->slug );
```
mboynes: Just pas the term object vs the slug. Passing the slug causes extra work, as WordPress has to then lookup the term again.

inc/components/class-menu.php
```diff
@@ -49,8 +61,55 @@ public function parse_wp_menu_by_location( string $menu_location ) {
...
+				if ( ! empty( $children_of_menu_item ) ) {
+					$clean_menu_item->set_children( $children_of_menu_item );
+				}
+
+				$menu[] = $clean_menu_item;
+			}
+		}
+
+		return (array) $menu;
+	}
```
mboynes: This all seems a bit unnecessary to me, though it's hard to tell without seeing the JS. It seems to me that this would double the net effort in building a menu: the menu items are looped-over in PHP, then inevitably JS will also have to loop over the items as well (just now in a different structure). Another approach would be to make a JS walker similar to WP's PHP walker, which would mean that the menu items are only looped-over once.

inc/components/class-menu.php
```diff
@@ -49,8 +61,55 @@ public function parse_wp_menu_by_location( string $menu_location ) {
...
+			$menu_item_id = $menu_item->ID;
+
+			// Current menu's parent id.
+			$menu_item_parent_id = absint( $menu_item->menu_item_parent );
+
+			// Is the current menu item a child of the parent item.
+			if ( $menu_item_parent_id === $parent_id ) {
+
+				// Get parsed menu item.
+				$clean_menu_item = menu_item()->parse_menu_post( $menu_item );
```
mboynes: if it's not rendered moot by my previous comment, writing functions with very generic names like `menu_item()` (even though it's namespaced) makes for confusing code. I had to dig around to see where that was defined, assuming it was a WP core function. Same comment applies to `menu()`

inc/components/class-menu.php
jameswburke: Max and I discussed this. Since the children menu items are passed via the `children` property Irving can automatically render them in the correct nested order. We could do it on the app side, but it's just as easy to do here.

inc/components/class-menu.php
jameswburke: It's a helper function for getting an instance of the menu_item component. I'm not sure what else would make sense here. I could shift to a pattern more like `( new \WP_Irving\Component\Menu_Item() )->function_call()` and there's nothing here that precludes that from being used, but I didn't see this being a blocker, so we moved forward with it. Happy to refactor moving forward if there's a strong argument to be made about

PR #10, opened 2018-06-27
Add support for Safe Redirect Manger

jameswburke:
* Adds integration to WP-Irving for Safe Redirect Manager Plugin.

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -109,10 +109,10 @@ class Components_Endpoint extends Endpoint {
 			$request
 		);
 
-		// Create the response object
+		// Create the response object.
 		$response = new \WP_REST_Response( $data );
 
-		// Add a custom status code
+		// Add a custom status code.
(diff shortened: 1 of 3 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -160,7 +200,7 @@ public function add_api_link( array $actions, \WP_Post $post ) : array {
 		$path_url = add_query_arg(
 			'path',
 			$path,
-			rest_url( 'irving/v1/components' )
+			site_url( 'wp-json/irving/v1/components' )
```
ethitter: Why this change?

jameswburke: `rest_url()` uses `WP_HOME` as the base of the URL, which in Irving is the app, so the URL being generated here wasn't actually pointing to the WP instance, but to Heroku.

ethitter: Ah, understood. I'd add a comment noting this. 

Also, rather than hard-coding `wp-json`, this can use `rest_get_url_prefix()`.

PR #21, opened 2018-07-26
Create HTML Component And Allow Configuring Images By ID

meddy:


inc/components/class-image.php
```diff
@@ -124,11 +124,22 @@ public static function register_crop_sizes( array $crop_sizes ) {
 	public function set_post_id( $post_id ) {
 		// Get the URL.
 		$attachment_id  = get_post_thumbnail_id( $post_id );
```
scottnelle: There are two spaces after `$attachment_id` and PHPCS is flagging it as a warning.

inc/components/class-content.php
```diff
@@ -68,11 +69,10 @@ private function map_block( array $block ) {
 		if ( ! empty( $block['innerHTML'] ) ) {
 			// Clean up extraneous whitespace characters.
 			$content = preg_replace( '/[\r\n\t\f\v]/', '', $block['innerHTML'] );
-			return new Component(
-				'rawHTML',
-				array_merge( $block['attrs'] ?? [], [ 'content' => $content ] ),
-				array_map( [ $this, 'map_block' ], $block['innerBlocks'] ?? [] )
-			);
+			return new Html( [
+				'config' => array_merge( $block['attrs'] ?? [], [ 'content' => $content ] ),
```
scottnelle: PHPCS is throwing a warning because this double arrow isn't aligned with the one below. Curiously, Travis is returning a failure on warnings for this project. Unsure if that's a configuration flaw or an intentional decision for Irving. Regardless, my night is now ruined because arrow alignment has caused a build to fail. ;-)

PR #26, opened 2018-08-01
WP-Irving term and byline

jameswburke:
* Setup Byline component (with support for CoAuthors Plus).
* Setup Term component.
* Fix PHPCS config.
* Only apply post row filters to published content.


inc/components/class-byline.php
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Define the default config of a byline.
+	 *
+	 * @return array Default config values for this component.
+	 */
+	public function default_config() {
+		return [
+			'avatar'       => null,
+			'display_name' => '',
```
meddy: use camelCase

inc/components/class-byline.php
```diff
@@ -0,0 +1,68 @@
...
+	public function set_coauthor( $coauthor ) {
+
+		// Validate object.
+		if ( 'guest-author' !== ( $coauthor->type ?? '' ) ) {
+			return null;
+		}
+
+		// Set byline configs.
+		$this->set_config( 'id', $coauthor->ID );
+		$this->set_config( 'display_name', $coauthor->display_name );
```
meddy: camelCase for the config key

inc/components/class-byline.php
```diff
@@ -0,0 +1,68 @@
...
+		$this->set_config( 'slug', $coauthor->user_login );
+		$this->set_config( 'link', get_author_posts_url( $coauthor->ID, $coauthor->user_nicename ) );
+
+		// Set the avatar image.
+		$avatar_id = get_post_thumbnail_id( $coauthor->ID );
+		if ( 0 !== absint( $avatar_id ) ) {
+			$this->set_config(
+				'avatar',
+				( new \WP_Irving\Component\Image() )
+					->set_attachment_id( $avatar_id )
```
meddy: You also have to set a size, I think, for the data to be configured.

jameswburke: We have to define that later since the media registration is done in the theme. We could probably setup some defaults in the plugin, but that'll have to happen later.

https://github.com/thriveglobal/thrive-global-wp/blob/production/themes/thrive-global/inc/components/class-byline.php#L53-L54

PR #30, opened 2018-08-11
Ensure camel case and whitelist config

jameswburke:
* Automatically camelCase all keys on output for API predictability.
* Adds a `$whitelist` array of config keys to limit output.


inc/components/class-component.php
```diff
@@ -147,11 +154,62 @@ public function set_name_of_children( string $name ) {
...
 	}
 
+	/**
+	 * Convert all array keys to camel case.
+	 *
+	 * @param array $array        Array to convert.
+	 * @param array $array_holder Parent array holder for recursive array.
+	 * @return array Updated array with camel-cased keys.
+	 */
+	public function camel_case_keys( $array, $array_holder = [] ) {
```
meddy: I can't understand the purpose of this 2nd argument.

PR #32, opened 2018-08-19
Bug/fix wp query setup

jameswburke:


inc/endpoints/class-components-endpoint.php
```diff
@@ -177,6 +177,23 @@ public function build_query( $path, $params ) {
...
+					$wp_rewrite->use_verbose_page_rules
+					&& preg_match( '/pagename=\$matches\[([0-9]+)\]/', $query, $varmatch )
+				) {
+					// This is a verbose page match, let's check to be sure about it.
+					$page = get_page_by_path( $matches[ $varmatch[1] ] );
+					if ( ! $page ) {
+				 		continue;
+					}
+
+					$post_status_obj = get_post_status_object( $page->post_status );
```
meddy: Maybe add a comment explaining the intention here.

inc/endpoints/class-components-endpoint.php
```diff
@@ -200,7 +217,13 @@ public function build_query( $path, $params ) {
 		// add_query_arg will encode the url, which we don't want.
 		$query = urldecode( $query );
 
-		return new \WP_Query( $query );
+		// Execute query.
+		$query = new \WP_Query( $query );
+
+		// Map to main query.
+		$wp_the_query = $query;
```
meddy: You should update `class-paginator.php`, and remove the global handling there, because you are doing it here.

Are you sure that completely replacing the global for the entire request won't have unforeseen side effects? I would think that kind of like reference counting, you would want to put the global back after you are done doing stuff. In this case it would be after building the components.

jameswburke: There's two global WP_Query objects, one is the `$wp_the_query`, which we're setting here (since REST requests don't execute one). This allows us to use `is_main()`. The paginator global is the instance of the WP_Query currently being looped through, so unfortunately, both are still needed.

PR #33, opened 2018-08-21
Handle not enough results to need pages.

meddy:


inc/components/class-paginator.php
```diff
@@ -48,7 +48,7 @@ public function set_from_wp_query( \WP_Query $wp_query ) {
 		// phpcs:ignore WordPress.WP.GlobalVariablesOverride
 		$GLOBALS['wp_query'] = $loop_query;
 
-		$this->set_children( array_map( [ $this, 'map_link' ], $links ) );
+		$this->set_children( array_map( [ $this, 'map_link' ], $links ?? [] ) );
```
mslinnea: Hmm, is $links actually null at this point?  Looking at the WordPress function `paginate_links()` it looks like it probably is returning an empty array...  But perhaps I'm missing something.  I guess I'm not seeing how this is helping.  Do you want to not add these children if $links is empty?

meddy: This appears to be the case where null may be returned.
https://github.com/WordPress/WordPress/blob/master/wp-includes/general-template.php#L3857

An empty array works here as a coalesced value, because the default value of children is an empty array, and mapping on empty array returns an empty array.

PR #34, opened 2018-08-22
Add an API link to all term row actions

jameswburke:


kevinfodness:
Looks like the Travis config for phpcs is busted, FYI.

PR #35, opened 2018-08-22
Allow retina, lazyload, and aspect_ratio to be configured per-size

ostowe:


inc/components/class-image.php
```diff
@@ -192,6 +192,11 @@ public function set_config_for_size( string $image_size, $picture = false ) {
 			$size_config = $sizes[ $image_size ];
 			$this->set_config( 'image_size', $image_size );
 			$this->set_config( 'sources', $size_config['sources'] );
+
+			// You can set certain config values on a per-size basis instead of per-component
+			$this->set_config( 'aspect_ratio', $size_config['aspect_ratio'] ?? $this->config['aspect_ratio'] );
+			$this->set_config( 'retina', $size_config['retina'] ?? $this->config['retina'] );
+			$this->set_config( 'lazyload', $size_config['lazyload'] ?? $this->config['lazyload'] );
```
jameswburke: Let's error proof this a bit more,
`$this->set_config( 'aspect_ratio', $size_config['aspect_ratio'] ?? $this->config['aspect_ratio'] ?? '' );`

That last `??`, just incase `$this->config['aspect_ratio']` is null too, we won't get a fatal error.

inc/components/class-image.php
```diff
@@ -364,7 +369,7 @@ public function get_srcset() : string {
 
 			// Add retina source to srcset, if applicable.
 			if ( is_numeric( $descriptor ) ) {
-				if ( $this->config['retina'] ) {
+				if ( $this->config['retina'] && ( empty( $params['retina'] ) || $params['retina'] ) ) {
```
jameswburke: `if ( $this->config['retina'] && empty( $params['retina'] ?? '' ) )`

Try this out, a little more verbose (I really love `??`)

PR #37, opened 2018-08-23
Update byline component to use gravatar as a fallback avatar image

ostowe:
- change sizing of global fallback image


inc/components/class-byline.php
```diff
@@ -60,6 +60,20 @@ public function set_coauthor( $coauthor, $avatar_size = 'avatar' ) {
...
+			$user = get_user_by( 'login', $coauthor->linked_account ?? '' );
+			// Use default gravatar fallback regardless of whether or not there's a linked user.
+			$gravatar_url = get_avatar_url(
+				$user->ID ?? $user->data->user_email ?? 0,
+				[
+					'size' => '300',
+				]
+			);
+
+			$this->children[] = ( new \WP_Irving\Component\Image() )
```
jameswburke: Wrap all this in a `if ( ! empty( $gravatar_url ) )` just in case.

PR #40, opened 2018-08-24
Updating head component

jameswburke:


inc/components/class-head.php
```diff
@@ -19,6 +19,22 @@ class Head extends Component {
...
+	 *
+	 * @param null|\WP_Query $wp_query WP_Query object or null.
+	 */
+	public function __construct( $wp_query = null ) {
+		parent::__construct();
+
+		// Set the default title.
+		$this->set_title( get_bloginfo( 'name' ) );
+
+		if ( $wp_query instanceof \WP_Query ) {
```
meddy: Every other component has kept the parent constructor interface, and then the query can be applied with a setter. Are you sure you want to break that consistency?

inc/components/class-head.php
```diff
@@ -28,11 +44,24 @@ public function default_config() {
...
 	}
 
+	/**
+	 * Set the state of this component using a $wp_query object.
+	 *
+	 * @todo  Clean up this function and make it more verbose.
+	 *
+	 * @param \WP_Query $wp_query WP_Query object.
+	 */
+	public function parse_from_query( $wp_query ) {
```
meddy: Other components have called this method `set_from_query`.

inc/components/class-head.php
```diff
@@ -28,11 +44,24 @@ public function default_config() {
...
+			$this->set_title( $wp_query->post->post_title );
+		}
+	}
+
 	/**
 	 * Set a title tag as a child of the Head component.
 	 *
 	 * @param string $value The title value.
-	 * @return Head
+	 * @return Head The instance of this component.
```
meddy: There can be multiple instances, so I'm not sure what **The** instance implies.

PR #42, opened 2018-08-31
Add additional hooks to modify the query

jameswburke:


inc/endpoints/class-components-endpoint.php
```diff
@@ -85,17 +99,18 @@ public function get_route_response( $request ) {
...
 		do_action( 'wp_irving_components_request', $request );
 
-		$params = $request->get_params();
+		$this->params = $request->get_params();
+
 		// Parse path and context.
-		$this->parse_path( $params['path'] ?? '' );
-		$this->context = $params['context'] ?? '';
+		$this->parse_path( $this->params['path'] ?? '' );
+		$this->context = $this->params['context'] ?? '';
```
pattiereaves: ooo those fancy operators

PR #44, opened 2018-09-04
Force trailing slashes

jameswburke:
https://alleyinteractive.atlassian.net/browse/IRV-95

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -112,6 +112,9 @@ class Components_Endpoint extends Endpoint {
 
 		$this->query = $this->build_query();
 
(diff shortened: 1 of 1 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -358,6 +361,34 @@ public function add_api_link_to_terms( array $actions, \WP_Term $term ) : array
...
+
+		// Apply all params to the rest url for this endpoint.
+		$request_url = add_query_arg(
+			$this->params,
+			rest_url( 'irving/v1/components' )
+		);
+
+		// Redirect permanently.
+		wp_redirect( $request_url, 301 );
+		exit();
```
meddy: Technically exit is a language construct, not a function, so you don't need parens.

PR #47, opened 2018-09-07
Update image component to better handle aspect ratios and lazy loading

ostowe:


inc/components/class-image.php
```diff
@@ -46,21 +46,24 @@ class Image extends Component {
...
 			'attachment_id' => 0,
-			'post_id'       => 0,
-			'image_size'    => 'full',
 			'alt'           => '',
 			'caption'       => '',
+			'crops'         => '',
+			'height'        => 0,
+			'image_size'    => 'full',
+			'lazyload'      => true,
+			'lqipSrc'      => '',
```
jameswburke: Alignment

PR #53, opened 2018-09-27
Add WP-Irving GA component

jameswburke:


inc/components/class-google-analytics.php
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @var string
+	 */
+	public $name = 'google-analytics';
+
+	/**
+	 * Initalize this component.
+	 */
+	public function __construct() {
+		parent::__construct();
```
meddy: I believe the parent constructor takes a few arguments, don't you want to pass those through?

jameswburke: Good point. I'll follow up in a subsequent PR.

PR #55, opened 2018-10-02
Add form endpoint class and filter for adding more form endpoints

ostowe:


inc/endpoints/class-form-endpoint.php
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	/**
+	 * Fix rest url.
+	 *
+	 * @see https://github.com/WordPress/gutenberg/issues/1761
+	 *
+	 * @param string $url Rest URL.
+	 */
+	public function fix_rest_url( $url ) {
```
ostowe: Not sure this is actually necessary here since we're already doing it for components endpoint

jameswburke: Safe to remove.

inc/endpoints/class-form-endpoint.php
```diff
@@ -0,0 +1,79 @@
...
+		 * @param array $form_endpoints {
+		 *     Form endpoint slugs and callback functions.
+		 *
+		 *     @type string $slug The slug for the form endpoint.
+		 *     @type string $callback response callback to use when the endpoint is called.
+		 * }
+		 */
+		$form_endpoints = (array) apply_filters(
+			'wp_irving_form_endpoints',
+			$this->form_endpoints
```
jameswburke: There's no reason to store as a class variable since you never set it. E.g. nowhere in the code does `$this->form_endpoints = something` exist. Might as well keep it all local for now.

PR #56, opened 2018-10-09
Updates to head component

jameswburke:
Additional head tag updates

### DIFF
--- a/inc/components/class-head.php
+++ b/inc/components/class-head.php
@@ -51,6 +51,16 @@ class Head extends Component {
 			$this->set_title( $wp_query->post->post_title );
 		}
 
+		// If queried object is a valid article post type.
+		$queried_object = $wp_query->get_queried_object();
+		if ( $queried_object instanceof \WP_Post ) {
+			$post_id = $queried_object->ID;
+
+			$this->apply_meta( $post_id );
+			$this->apply_social_meta( $post_id );
+			$this->set_title( $this->get_meta_title( $post_id ) );
+		}
+
 		return $this;
 	}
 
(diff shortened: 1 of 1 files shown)


inc/components/class-head.php
```diff
@@ -170,4 +180,154 @@ protected function add_tag( $tag, $attributes = [] ) {
...
+	 */
+	public function get_meta_description( $post_id ) : string {
+
+		// Return WP-SEO description.
+		$meta_description = (string) get_post_meta( $post_id, '_meta_description', true );
+		if ( ! empty( $meta_description ) ) {
+			return $meta_description;
+		}
+
+		// Modify global state.
```
tomharrigan: The following should accomplish the same without modifying global state and having to set the temp variable `apply_filters( 'get_the_excerpt', get_post_field( 'post_excerpt', $post_id ) )`

inc/components/class-head.php
```diff
@@ -170,4 +180,154 @@ protected function add_tag( $tag, $attributes = [] ) {
...
+	 * Get the social image for a post.
+	 *
+	 * @param  integer $post_id     Post ID.
+	 * @param  array   $photon_args Photon API arguments.
+	 * @return string Social image url.
+	 */
+	public function get_social_image_url( $post_id, $photon_args = [] ) {
+
+		// Get image url.
+		$image_id = absint( get_post_meta( $post_id, 'social_image_id', true ) );
```
tomharrigan: might be good to return early if this isn't valid to avoid a `wp_get_attachment_image_url` call

inc/components/class-head.php
jameswburke: I had a long discussion with @mboynes about this and this was his recommended approach. He suggested avoiding the filter all together.

PR #62, opened 2018-10-24
Fix encoding for whatsapp sharing

ostowe:
Refs https://alleyinteractive.atlassian.net/browse/TG-90

### DIFF
--- a/inc/components/class-social-share.php
+++ b/inc/components/class-social-share.php
@@ -30,6 +30,7 @@ class Social_Share extends Component {
 		return [
 			'post_id'        => 0,
 			'display_icons'  => true,
+			'cta_template'   => 'Check out this story: %1s %2s',
 		];
 	}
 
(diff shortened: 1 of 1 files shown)


inc/components/class-social-share.php
```diff
@@ -141,11 +141,16 @@ public function get_excerpt( $post ) {
...
-		return urlencode( sprintf(
-			'Check out this story: %1s from Thrive global %2s',
-			$this->get_title( $post ),
-			$this->get_url( $post )
-		) );
+		return rawurlencode(
+			// Decode HTML entities first or we might end up with some lingering in the share text.
+			html_entity_decode(
+				sprintf(
+					'Check out this story, %1s from Thrive global: %2s',
```
amyevans: We shouldn't reference Thrive here, also should use `esc_html__()`

PR #66, opened 2018-11-02
Code Fixes for 404 Changes

jvangemeren:


inc/endpoints/class-components-endpoint.php
```diff
@@ -236,9 +239,12 @@ public function build_query() {
 					// This is a verbose page match, let's check to be sure about it.
 					$page = get_page_by_path( $matches[ $varmatch[1] ], OBJECT, $page_type );
 					if ( ! $page ) {
+						$is_404 = true;
 						continue;
 					}
 
+					$is_404 = false;
```
GFargo: I could be missing something or not following indentation correctly, but if we default this to `false` on line `219` do we need to set it here too?

GFargo: Ah but this _would_ be required if we're flipping back and forth during the `foreach` loop... I'll leave it up to you 😸

inc/components/class-head.php
```diff
@@ -41,18 +41,22 @@ public function set_from_query( $wp_query ) {
...
-		if ( $queried_object instanceof \WP_Post ) {
-			$post_id = $queried_object->ID;
-
-			$this->apply_meta( $post_id );
-			$this->apply_social_meta( $post_id );
-			$this->set_title( $this->get_meta_title( $post_id ) );
-		} elseif ( $queried_object instanceof \WP_Term ) {
-			$this->set_title( $queried_object->name );
+		if( $wp_query->is_404() ) {
+			$this->set_title( '404 - Page not found' );
```
amyevans: You'll also want any text strings to be translated:
```suggestion
			$this->set_title( __( '404 - Page not found', 'wp-irving' ) );
```

PR #70, opened 2018-11-06
Remove esc_url from lqipSrc as it is escaping `&` separating query vars

ostowe:


inc/components/class-image.php
```diff
@@ -237,7 +237,7 @@ public function configure( $picture ) {
 			'alt'         => $this->get_alt_text(),
 			'caption'     => ! empty( $this->config['attachment_id'] ) ? wp_get_attachment_caption( $this->config['attachment_id'] ) : '',
 			'height'      => $image_meta['height'] ?? 0,
-			'lqipSrc'     => esc_url( $this->get_lqip_src()->config['url'] ),
```
mslinnea: @ostowe esc_url_raw should fix this

PR #71, opened 2018-11-06
Return strin from get_lqip_src instead of class

ostowe:
Otherwise we can't know if it failed to retrieve the lqip src and will just end up using the full size image...which totally defeats the purpose :/

### DIFF
--- a/inc/components/class-image.php
+++ b/inc/components/class-image.php
@@ -237,7 +237,7 @@ class Image extends Component {
 			'alt'         => $this->get_alt_text(),
 			'caption'     => ! empty( $this->config['attachment_id'] ) ? wp_get_attachment_caption( $this->config['attachment_id'] ) : '',
 			'height'      => $image_meta['height'] ?? 0,
-			'lqipSrc'     => $this->get_lqip_src()->config['url'],
+			'lqipSrc'     => $this->get_lqip_src(),
 			'originalUrl' => $this->get_base_url(),
 			'picture'     => $picture,
 			'sizes'       => $this->get_sizes(),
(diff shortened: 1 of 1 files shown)


inc/components/class-image.php
```diff
@@ -257,7 +257,9 @@ public function configure( $picture ) {
 	 * @return string
 	 */
 	public function get_aspect_ratio( $size_config ) {
-		$aspect_ratio = $size_config['aspect_ratio'] ?? $this->config['aspect_ratio'] ?? false;
+		$aspect_ratio = isset( $size_config['aspect_ratio'] ) ?
```
GFargo: This should work... if you wanted to shorten, not sure it's 'simplification' ;)

```suggestion
		$aspect_ratio = ( $size_config['aspect_ratio'] ?? $this->config['aspect_ratio'] ) ?? false;
```

inc/components/class-image.php
```diff
@@ -257,7 +257,9 @@ public function configure( $picture ) {
 	 * @return string
 	 */
 	public function get_aspect_ratio( $size_config ) {
```
GFargo: Could add some nullable typehinting to the return value here
```suggestion
	public function get_aspect_ratio( $size_config ): ?string  {
```

inc/components/class-image.php
```diff
@@ -377,13 +379,13 @@ public function get_lqip_src() {
 
 		// Return early if no aspect ratio is set.
 		if ( empty( $aspect_ratio ) ) {
-			return $this;
+			return '';
```
GFargo: If you type-hinted this function, per comment above you'd want to update this to be `null`
```suggestion
			return null;
```

inc/components/class-image.php
ostowe: oo didn't know about these, nice!

PR #75, opened 2018-12-06
Parsely integration

ostowe:
Class for setting up and managing Parsely analytics and metadata.

### DIFF
--- a/inc/components/class-parsely.php
+++ b/inc/components/class-parsely.php
@@ -33,7 +33,7 @@ class Parsely extends Component {
 	 */
 	public function default_config() {
 		return [
-			'site'        => $this->get_parsely_site(),
+			'site' => $this->get_parsely_site(),
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


inc/components/class-parsely.php
```diff
@@ -0,0 +1,162 @@
...
+	 * Helper function that returns all the attributes needed for Parsely on a
+	 * single article.
+	 *
+	 * @param  \WP_Post $post Post object.
+	 * @return array
+	 */
+	public function get_parsely_meta( \WP_Post $post ) : array {
+
+		// Base properties.
+		$meta['parsely-link']    = get_the_permalink( $post );
```
pattiereaves: just curious, why declare it this way instead of `$meta = [ 'parsely-link' => /* val*/, /* ... etc. */ ]`

ostowe: Good question, I copypasta'd this code and mostly focused on removing unnecessary things. I agree with you though, will fix!

PR #78, opened 2019-01-31
Add helpers for common form endpoint responses

ostowe:


inc/endpoints/class-form-endpoint.php
```diff
@@ -51,6 +51,41 @@ public function register_rest_routes() {
 			);
 		}
 	}
+
+
+	/**
+	 * Generic internal server error response.
+	 *
+	 * @return WP_REST_Response
```
mslinnea: This should be `\WP_REST_Response`  This should be changed for all of the phpdoc in this pull request.  Leaving the slash off will prevent code completion from working correctly and some IDEs like phpstorm will highlight it as an error.

PR #89, opened 2019-04-17
Add base integration for Google AMP plugin

amyevans:


inc/integrations/class-google-amp.php
```diff
@@ -0,0 +1,43 @@
...
+	public function amp_template_redirect() {
+		if ( is_amp_endpoint() ) {
+			amp_render();
+		}
+	}
+}
+
+add_action( 'init', function() {
+	new \WP_Irving\Google_AMP();
+} );
```
jameswburke: Being picky, but I've been breaking this up into lines,

```
add_action(
	'init',
	function() {
		new \WP_Irving\Google_AMP();
	}
);
```

jameswburke: I agree that it's kinda ridiculously long, but it's what I've seen PHPCS prefer.

PR #97, opened 2019-07-22
Add tests for safe redirect manager and legacy redirector integrations

ostowe:
Refs https://alleyinteractive.atlassian.net/browse/IRV-135


inc/integrations/class-wpcom-legacy-redirector.php
```diff
@@ -48,13 +48,15 @@ public function handle_redirect( $data, $query, $context, $path, $request ) : ar
 		}
 
 		// Get the path parameter.
-		$request_path = apply_filters( 'wpcom_legacy_redirector_request_path', $params['path'] );
+		$request_path = trailingslashit( apply_filters( 'wpcom_legacy_redirector_request_path', $params['path'] ) );
```
renatonascalves: If someone filters it and adds the trailing slash, will this add another?!

ostowe: No, `trailingslashit` prevents two from being added (it'll remove any existing trailing slash first)

PR #101, opened 2019-09-05
Purging post cache

renatonascalves:
Purging the cache automatically on post `save/delete/update`.


inc/class-purge-cache.php
```diff
@@ -0,0 +1,164 @@
...
+				<?php esc_html_e( 'WP-Irving - Site Cache', 'wp-irving' ); ?>
+			</h1>
+
+			<hr class="wp-header-end">
+
+			<form method="post">
+				<?php settings_fields( $this->id ); ?>
+
+				<h3>
+					<?php esc_attr_e( 'Clear Site Cache', 'wp-irving' ); ?>
```
jacobschweitzer: Since this isn't an attribute it is probably more appropriate to use `esc_html_e`

inc/class-purge-cache.php
```diff
@@ -0,0 +1,164 @@
...
+				</h3>
+
+				<p>
+					<?php esc_attr_e( 'Use with care. Clearing the entire site cache will negatively impact performance for a short period of time.', 'wp-irving' ); ?>
```
jacobschweitzer: Same here - `esc_html_e`

inc/class-purge-cache.php
```diff
@@ -0,0 +1,164 @@
...
+				</p>
+
+				<p class="submit">
+					<input type="submit" name="submit" id="submit" class="button" value="<?php echo esc_attr( 'Clear Cache', 'wp-irving' ); ?>">
```
jacobschweitzer: `esc_attr` only accepts one argument, I think it should be `esc_attr_e`
```suggestion
					<input type="submit" name="submit" id="submit" class="button" value="<?php echo esc_attr_e( 'Clear Cache', 'wp-irving' ); ?>">
```

inc/class-purge-cache.php
renatonascalves:
```suggestion
					<?php esc_html_e( 'Clear Site Cache', 'wp-irving' ); ?>
```

inc/class-purge-cache.php
renatonascalves:
```suggestion
					<?php esc_html_e( 'Use with care. Clearing the entire site cache will negatively impact performance for a short period of time.', 'wp-irving' ); ?>
```

inc/class-purge-cache.php
```diff
@@ -0,0 +1,164 @@
...
+
+		// Fire the request.
+		wp_remote_get( $request_url );
+	}
+
+	/**
+	 * Fire wipe out request.
+	 */
+	protected function fire_wipe_request() : void {
+		wp_remote_get( '/bust-entire-cache' );
```
renatonascalves:
```suggestion
		wp_remote_get(  home( '/bust-entire-cache' ) );
```

renatonascalves:
Ready for another round of reviews! :)

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+		$this->fire_purge_request( $post_id );
+	}
+
+	/**
+	 * Fire purge request.
+	 *
+	 * @param int    $post_id   Post ID.
+	 * @param string $permalink Permalink.
+	 */
+	protected function fire_purge_request( $post_id = 0, $permalink = '' ) {
```
dlh01: The signature of this method is a little confusing in that the `$permalink` parameter silently overrides the `$post_id` parameter. Could this function accept just the URL, without being aware of posts or any object type? When posts need to be purged, their permalink could be passed directly.

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+
+		// Get post permalink with fallback.
+		if ( empty( $permalink ) ) {
+			$permalink = get_the_permalink( $post_id );
+		}
```
dlh01: - If `$post_id` is `0`, this would purge the global `$post`, which might be unexpected. 
- An invalid post ID will return `false`. What would that purge?
- What if the post is now in the trash? The permalink will include the `__trashed` suffix. Do the cache implementations know how to handle those?

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+	/**
+	 * Class constructor.
+	 */
+	public function __construct() {
+		// Register admin page.
+		add_action( 'admin_menu', [ $this, 'register_admin' ] );
+
+		// Purging actions.
+		add_action( 'wp_insert_post', [ $this, 'on_update_post' ], 10, 2 );
+		add_action( 'transition_post_status', [ $this, 'on_post_status_transition' ], 10, 3 );
```
dlh01: - Is there a situation where `wp_insert_post` would fire but `transition_post_status` would not fire? Only the `transition_post_status` hook might be necessary.
- Also, does the cache need to be purged when non-viewable post types change status?

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	/**
+	 * Purge cache on a PURGE request.
+	 */
+	public function purge_cache_request() {
+		global $wp;
+
+		if ( isset( $wp->request ) && 'PURGE' === strtoupper( $_SERVER['REQUEST_METHOD'] ) ) { // phpcs:ignore
+			$this->fire_purge_request( 0, home_url( $wp->request ) );
```
dlh01: Is there any sort of authentication on who can send a request to WordPress that will cause `init` to fire? Can anyone send a `PURGE` request?

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+			<h1 class="wp-heading-inline">
+				<?php esc_html_e( 'WP-Irving - Site Cache', 'wp-irving' ); ?>
+			</h1>
+
+			<hr class="wp-header-end">
+
+			<form method="post">
+				<?php settings_fields( 'irving-cache' ); ?>
+
+				<h3>
```
dlh01: Why skip `<h2>`?

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+					<?php esc_html_e( 'Clear Site Cache', 'wp-irving' ); ?>
+				</h3>
+
+				<p>
+					<?php esc_html_e( 'Use with care. Clearing the entire site cache will negatively impact performance for a short period of time.', 'wp-irving' ); ?>
+				</p>
+
+				<p class="submit">
+					<input type="submit" name="submit" id="submit" class="button" value="<?php echo esc_attr_e( 'Clear Cache', 'wp-irving' ); ?>">
+				</p>
```
dlh01: Can `\submit_button()` be used?

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
...
+			</form>
+
+			<hr>
+		</div>
+		<?php
+	}
+}
+
+new \WP_Irving\Purge_Cache();
```
dlh01: Can this be instantiated outside of the class file so that including the file doesn't automatically load the purging logic? Because there is no reference to the class instance available, once it's on, it's pretty tough to turn off.

inc/class-purge-cache.php
```diff
@@ -0,0 +1,183 @@
+<?php
+/**
+ * WP Irving Cache.
```
dlh01:
```suggestion
 * WP Irving Cache
```
No periods in file headers, [per the core standards](https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#6-file-headers).

inc/class-purge-cache.php
renatonascalves: No need! Anyone can request via the `PURGE` method and the cache will be purged. This is for VIP.

renatonascalves:
@dlh01 Can you take another look?

PR #102, opened 2019-10-01
Fix the API endpoint URL for multisites with subdirectories

amyevans:
* Modify the logic for getting the path of a URL (use `site_url` instead of `wp_parse_url`). Fixes a bug where the API endpoint for a multisite with subdirectories was incorrect.
* Pull duplicate logic into a helper function

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -524,6 +524,10 @@ class Components_Endpoint extends Endpoint {
 			return;
 		}
 
+		if ( ! function_exists( 'get_current_screen' ) ) {
+			return;
+		}
+
 		// Get screen and check for a post base.
 		$screen = get_current_screen();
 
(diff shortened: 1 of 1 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -453,26 +473,15 @@ public function add_api_link_to_posts( array $actions, \WP_Post $post ) : array
...
-		// WP_SITEURL, which is what we need to use.
-		$path_url = add_query_arg(
-			'path',
-			$path,
-			site_url( rest_get_url_prefix() . '/irving/v1/components' )
-		);
-
-		$path_url = apply_filters( 'wp_irving_post_row_action_path_url', $path_url, $post );
+		// Get the API URL, allowing it to be filtered.
+		$api_url = self::get_wp_irving_api_url( $permalink );
```
kevinfodness: Since you're actually getting the API URL for a specific path, I'd leave the variable named `$path_url` for clarity.

PR #104, opened 2019-10-04
Register rewrite rules for REST routes

amyevans:
Register rewrite rules so that rest routes for multisites with subdirectories work on decoupled sites

### DIFF
--- a/inc/rewrites.php
+++ b/inc/rewrites.php
@@ -15,19 +15,15 @@ namespace WP_Irving;
  */
 function add_additional_rest_routes( $rules ) {
 
-	// Nothing special needed for single sites.
-	if ( ! is_multisite() ) {
-		return $rules;
-	}
-
 	$site_paths = get_site_paths();
+	$prefix     = rest_get_url_prefix();
 
 	foreach ( $site_paths as $path ) {
 		$rest_api = [
-			"^{$path}/wp-json/?$"              => 'index.php?rest_route=/',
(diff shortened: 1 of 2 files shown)


inc/rewrites.php
```diff
@@ -0,0 +1,56 @@
...
+}
+add_filter( 'rewrite_rules_array', __NAMESPACE__ . '\add_additional_rest_routes' );
+
+/**
+ * Get all site paths for a multisite.
+ *
+ * @return array
+ */
+function get_site_paths() {
+	$sites = get_sites();
```
dlh01: Because `get_sites()` won't exist in a single-site installation, and so calling this function would generate a fatal error, perhaps the `is_multisite()` check should happen here instead of in `add_additional_rest_routes()`, with an empty array returned straightaway if it's a single-site installation. `add_additional_rest_routes()` would then do a `foreach()` over nothing and be a no-op.

inc/rewrites.php
```diff
@@ -0,0 +1,56 @@
...
+	// Nothing special needed for single sites.
+	if ( ! is_multisite() ) {
+		return $rules;
+	}
+
+	$site_paths = get_site_paths();
+
+	foreach ( $site_paths as $path ) {
+		$rest_api = [
+			"^{$path}/wp-json/?$"              => 'index.php?rest_route=/',
```
dlh01: Consider using `rest_get_url_prefix()` instead of `wp-json`.

PR #106, opened 2019-10-05
Add helper for debugging the WP_Query and a default welcome message

amyevans:
This PR:
* hooks into `wp_irving_components_route` at priority 100, and if a theme has set `wp_irving_wp_query_debug` to return `true`, spits out the WP_Query as an HTML component for quicker debugging
*  hooks into `wp_irving_components_route` at priority 1000, and if all of the values that could've/should've been modified by a theme are still empty, prints a welcome message

I'm somewhat on the fence about the welcome message, so if you have an opinion let me know.


inc/debug.php
```diff
@@ -0,0 +1,148 @@
...
+	\WP_REST_Request $request
+): array {
+
+	if (
+		empty( $data['defaults'] )
+		&& empty( $data['page'] )
+		&& empty( $data['providers'] )
+		&& '' === $data['redirectTo']
+		&& 0 === $data['redirectStatus']
+	) {
```
jameswburke: I'd like to see this conditional reversed. Exit early, and exit often.

PR #107, opened 2019-10-08
Yoast Webmaster Tools integration

renatonascalves:


inc/integrations/class-wordpress-seo.php
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @package WP_Irving;
+ */
+
+namespace WP_Irving;
+
+/**
+ * Class to add WordPress_SEO's query modifications.
+ */
+class WordPress_SEO {
```
zrothauser: This looks good, the code in the constructor and the `get_webmaster_tools_codes()` method should be able to be taken from here and put into my `Yoast` class once #108 is merged in.

renatonascalves:
@zrothauser Can you take another look?

PR #109, opened 2019-10-13
Fix for Safe Redirect Manager

renatonascalves:
Sites that have the Safe Redirect Manager plugin activated but without this part/PR, https://github.com/10up/safe-redirect-manager/pull/176, applied fails. Meaning the endpoint fails because the method doesn't exist yet.

This PR fixes the fatal error. As soon as the PR is applied to the plugin, we can remove it.

### DIFF
--- a/tests/test-safe-redirect-manager.php
+++ b/tests/test-safe-redirect-manager.php
@@ -31,10 +31,22 @@ class Safe_Redirect_Manager_Tests extends WP_UnitTestCase {
 		self::$helpers = new \WP_Irving\Test_Helpers();
 	}
 
+	public function setUp() {
+		parent::setUp();
+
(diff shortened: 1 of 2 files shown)


tests/test-safe-redirect-manager.php
```diff
@@ -31,10 +31,22 @@ public static function setUpBeforeClass() {
...
+		parent::setUp();
+
+		$this->object = \SRM_Redirect::factory();
+	}
+
 	/**
 	 * Test relative redirect.
 	 */
 	public function test_relative_to_relative() {
+		if ( ! method_exists( $this->object, 'get_redirect_match' ) ) {
```
ostowe: could you abstract this function so you don't have to keep repeating it for each test method?

PR #110, opened 2019-10-13
User auth: login endpoint

renatonascalves:
This endpoint is reponsible for logging in the user and creating the necessary token which will be used on the React/Node/irving site.

The implementation updates the API Key-pairs with the generated keys. Meaning the tokens created here are reflected on the user profile list of api keys.

And it requires the https://github.com/WP-API/jwt-auth plugin to be installed and active.

### DIFF
--- a/inc/endpoints/class-login-endpoint.php
+++ b/inc/endpoints/class-login-endpoint.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * Class file for login endpoints.
+ *
+ * @package WP_Irving
+ */
+
+namespace WP_Irving\REST_API;
+
+/**
+ * Login Endpoint.
+ */
+class Login_Endpoint extends Endpoint {
+
+	/**
+	 * Attach to required hooks for login endpoint
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		add_action( 'rest_api_init', [ $this, 'register_rest_routes' ] );
+	}
+
+	/**
+	 * Register the REST API routes.
+	 */
+	public function register_rest_routes() {
+		register_rest_route(
+			self::get_namespace(),
+			'/login/',
+			[
+				'methods'  => 'GET',
+				'callback' => [ $this, 'get_route_response' ],
+				'args'     => array(
+					'username'    => array(
+						'description'       => esc_html__( 'The user name or email address.', 'wp-irving' ),
+						'type'              => 'string',
+						'required'          => true,
+						'sanitize_callback' => 'sanitize_text_field',
+						'validate_callback' => 'rest_validate_request_arg',
+					),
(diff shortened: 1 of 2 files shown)


inc/endpoints/class-login-endpoint.php
```diff
@@ -0,0 +1,116 @@
...
+	public function get_route_response( $request ) {
+
+		// JWT Auth plugin needs to be installed and active.
+		if ( ! class_exists( '\WP_REST_Key_Pair' ) ) {
+			return new \WP_Error(
+				'rest_authentication_jwt_auth_plugin_required',
+				__( 'The JWT Auth needs to be installed and active for this endpoint to work correctly.', 'wp-irving' ),
+				array(
+					'status' => 500,
+				)
```
jameswburke: Short array syntax.

inc/endpoints/class-login-endpoint.php
```diff
@@ -0,0 +1,116 @@
...
+			);
+		}
+
+		// Authenticate user first.
+		$user = wp_authenticate( $request['username'], $request['password'] );
+		if ( false === $user || ! ( $user instanceof \WP_User ) ) {
+			return new \WP_Error(
+				'rest_authentication_invalid_user',
+				__( 'There was a problem authenticating the user.', 'wp-irving' ),
+				array(
```
jameswburke: Short array syntax.

inc/endpoints/class-login-endpoint.php
```diff
@@ -0,0 +1,116 @@
...
+					'status' => 403,
+				)
+			);
+		}
+
+		$user_id       = $user->ID;
+		$api_key       = $user_id . wp_generate_password( 24, false );
+		$api_secret    = wp_generate_password( 32 );
+		$hashed_secret = wp_hash( $api_secret );
+		$new_item      = [
```
jameswburke: Can we come up with a better name than `$new_item`? I don't know what this is just from the name alone.

inc/endpoints/class-login-endpoint.php
```diff
@@ -0,0 +1,116 @@
...
+
+		$token_request = new \WP_REST_Request( \WP_REST_Server::CREATABLE, '/wp/v2/token' );
+		$token_request->set_query_params(
+			[
+				'api_key'    => $api_key,
+				'api_secret' => $api_secret,
+			]
+		);
+
+		return rest_do_request( $token_request );
```
jameswburke: All of this would benefit from some documentation via comments. It's not obvious from just the code what's going on, especially if you're not familiar with the plugin/web tokens/key pairs.

renatonascalves:
@jameswburke Updated the code with more information to what's going on.

PR #131, opened 2020-04-09
Use core WP's query var allow-list

jameswburke:
Filter the component request parameters using core's allow-list.

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -132,6 +132,7 @@ class Components_Endpoint extends Endpoint {
 	 */
 	public function get_route_response( $request ) {
 
+		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.OverrideProhibited
 		global $wp;
 
 		/**
@@ -145,7 +146,7 @@ class Components_Endpoint extends Endpoint {
 		// global $wp object's $public_query_vars array.
 		$this->params = array_intersect_key(
 			$request->get_params(),
-			array_flip( $wp->public_query_vars ),
+			array_flip( $wp->public_query_vars )
 		);
 
 		// Parse path and context.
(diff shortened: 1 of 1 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -131,14 +131,23 @@ public function register_rest_routes() {
...
 		 * @param \WP_REST_Request $request  WP_REST_Request object.
 		 */
 		do_action( 'wp_irving_components_request', $request );
 
-		$this->params = $request->get_params();
+		// Remove any request parameters that haven't been allow-listed by the
+		// global $wp object's $public_query_vars array.
+		$this->params = array_intersect_key(
+			$request->get_params(),
+			array_flip( $wp->public_query_vars )
```
pattiereaves, line 149: oh that is clever

joemcgill: Here's the full list of where WP adds public query vars:
https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/class-wp.php#L9-L17

Are we sure we want/need to support all of these explicitly? Many of them should be handled by the route parsing on the WP side, I would think.

inc/endpoints/class-components-endpoint.php
```diff
@@ -366,7 +375,7 @@ public function build_query() {
 		$wp_query = apply_filters( 'wp_irving_components_wp_query', $wp_query, $this->path, $this->custom_params, $this->params );
 
 		// Map to main query and set up globals.
-		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited
+		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited, WordPress.WP.GlobalVariablesOverride.OverrideProhibited
```
joemcgill, line 378: Did the sniff rule change, or do we really need to be ignoring both here?

jameswburke: I'm not sure I know enough about each of those vars to make an educated decision on whether or not they should be present. I do believe we should be mimicking WordPress core as much as possible though.

One of those phpcs rules hits me locally, one hits me on Travis, so I included both for the sake of it right now.

jameswburke:
We'll continue this discussion in #132

PR #133, opened 2020-04-11
Feature/irv 308/jwt auth

jameswburke:
Adds a new integration for the WP-API/JWT auth plugin.

This creates a new Bearer token to be used in Authentication headers, and is cross-domain, so Irving core can use it.

See https://github.com/alleyinteractive/irving/tree/feature/IRV-313/components-api-auth for the companion update in Irving core.

### DIFF
--- a/inc/class-previews.php
+++ b/inc/class-previews.php
@@ -47,7 +47,7 @@ class Previews {
 		remove_action( 'init', '_show_post_preview' );
 
 		// Filter to enable/disable public previews.
-		if ( apply_filters( 'wp_irving_enable_public_previews', true ) ) {
+		if ( apply_filters( 'wp_irving_enable_public_previews', false ) ) {
 			add_action( 'wp_irving_components_request', [ $this, 'enabled_public_previews' ] );
 		}
 
--- a/inc/class-previews.php
+++ b/inc/class-previews.php
@@ -135,7 +135,7 @@ class Previews {
 			return $wp_query;
 		}
 
-		// Get all the revisions
+		// Get all the revisions.
 		$revisions = wp_get_post_revisions( $wp_query->post->ID );
 
 		// Use the preview_id, or fallback to the most recent revision.
--- a/inc/class-previews.php
+++ b/inc/class-previews.php
@@ -122,12 +122,6 @@ class Previews {
 	 */
 	public function modify_wp_query_for_previews( $wp_query, $path, $custom_params, $params ) {
 
-		// Disable revision preview nonces so we can inject our own logic.
-		// Without this, wp_die() will execute, which means our auth strategy
-		// fails. There isn't anyway to currently filter this behavior, so
-		// we're stuck with this.
-		remove_action( 'init', '_show_post_preview' );
-
 		// Auth and a real route required.
 		if ( ! is_user_logged_in() || ! $wp_query->have_posts() ) {
 			return $wp_query;
@@ -136,19 +130,16 @@ class Previews {
 		$is_preview = wp_validate_boolean( $params['preview'] ?? false );
 		$preview_id = absint( $params['preview_id'] ?? 0 );
 
-		// This is not a preview.
-		if (
-			false === $is_preview
-			|| 0 === $preview_id
-		) {
+		// Require `preview` to be true, and `p` to be an integer.
+		if ( false === $is_preview || 0 === $preview_id ) {
 			return $wp_query;
 		}
 
-		if ( ( $wp_query->post->ID ?? 0 ) === $preview_id ) {
-			$revision = current( wp_get_post_revisions( $preview_id ) );
-		} else {
-			$revision = wp_get_post_revision( $preview_id );
-		}
+		// Get all the revisions
+		$revisions = wp_get_post_revisions( $wp_query->post->ID );
+
+		// Use the preview_id, or fallback to the most recent revision.
+		$revision = $revisions[ $preview_id ] ?? current( $revisions );
 
 		$wp_query->posts[0] = $revision;
 		$wp_query->post     = $revision;
--- a/inc/class-previews.php
+++ b/inc/class-previews.php
@@ -42,18 +42,25 @@ class Previews {
 		add_filter( 'query_vars', [ $this, 'modify_query_vars' ] );
 
 		// Filter to easily enable public previews.
-		if ( apply_filters( 'wp_irving_enable_public_previews', '__return_false' ) ) {
+		if ( apply_filters( 'wp_irving_enable_public_previews', false ) ) {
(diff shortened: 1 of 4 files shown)


inc/integrations/class-jwt-auth.php
```diff
@@ -0,0 +1,157 @@
...
+		}
+		return static::$instance;
+	}
+
+	/**
+	 * Setup the singleton.
+	 *
+	 * @todo Determine if we need to validate the token. This may not be needed
+	 * because the cookie is set to expire right before the auth token does.
+	 * Maybe we should hook into if/when the token is revoked so we can kill
```
renatonascalves, line 60: This is the hook you can use to do that: https://github.com/WP-API/jwt-auth/blob/ca76a9cd3e2a9bb5524188f710ab2ee959d7cd28/wp-includes/rest-api/auth/class-wp-rest-token.php#L406

This [hook](https://github.com/WP-API/jwt-auth/blob/develop/wp-includes/rest-api/auth/class-wp-rest-key-pair.php#L290-L307) is the one the plugin is using to revoke the token.

jameswburke: Thanks - worked like a charm.

inc/class-previews.php
```diff
@@ -0,0 +1,153 @@
...
+	/**
+	 * Setup the singleton. Validate JWT is installed, and setup hooks.
+	 */
+	public function setup() {
+
+		// Allow-list some query vars.
+		add_filter( 'query_vars', [ $this, 'modify_query_vars' ] );
+
+		// Filter to easily enable public previews.
+		if ( apply_filters( 'wp_irving_enable_public_previews', '__return_false' ) ) {
```
amyevans, line 45:
```suggestion
		if ( apply_filters( 'wp_irving_enable_public_previews', false ) ) {
```

inc/class-previews.php
```diff
@@ -0,0 +1,153 @@
...
+			$this->enable_public_previews();
+		}
+
+		// Validate that JWT exists and is enabled.
+		if ( ! defined( 'JWT_AUTH_VERSION' ) ) {
```
amyevans, line 50: This check happens too early, at least in my experience. Perhaps move the creation of the class instance to the `plugins_loaded` hook like you've done in the JWT_Auth class.

inc/class-previews.php
```diff
@@ -0,0 +1,153 @@
...
+	 * @param string    $params        Request params.
+	 * @return \WP_Qury $wp_query WP_Query object for the request.
+	 */
+	public function modify_wp_query_for_previews( $wp_query, $path, $custom_params, $params ) {
+
+		// Disable revision preview nonces so we can inject our own logic.
+		// Without this, wp_die() will execute, which means our auth strategy
+		// fails. There isn't anyway to currently filter this behavior, so
+		// we're stuck with this.
+		remove_action( 'init', '_show_post_preview' );
```
amyevans, line 122: This is happening too late, I've already hit the `wp_die()` message of `Sorry, you are not allowed to preview drafts` before it gets here to remove that action. I'd move this line to the setup function since it's used in both cases (authed previews and public previews).

inc/class-previews.php
```diff
@@ -0,0 +1,153 @@
...
+		// This is not a preview.
+		if (
+			false === $is_preview
+			|| 0 === $preview_id
+		) {
+			return $wp_query;
+		}
+
+		if ( ( $wp_query->post->ID ?? 0 ) === $preview_id ) {
+			$revision = array_shift( wp_get_post_revisions( $preview_id ) );
```
amyevans, line 141:
```suggestion
			$revision = current( wp_get_post_revisions( $preview_id ) );
```

inc/class-previews.php
jameswburke: Good call. Moved into the action.

inc/class-previews.php
jameswburke: Also a good call. Updated.

jameswburke:
@amyevans, I updated the logic on previews a bit. Let me know if it looks good to you. I'm considering applying the revision logic across the board (so it applies when public previews are on too).

inc/class-previews.php
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		// Get all the revisions.
+		$revisions = wp_get_post_revisions( $wp_query->post->ID );
+
+		// Use the preview_id, or fallback to the most recent revision.
+		$revision = $revisions[ $preview_id ] ?? current( $revisions );
+
+		$wp_query->posts[0] = $revision;
+		$wp_query->post     = $revision;
```
jameswburke, line 145: Actually, instead of whole-sale replacing the post, I think we should only replace the title, content, and excerpt, since those are the only things that are allow-listed as revision data in WP Core. In testing, I'm seeing meta data being dropped b/c it's associated with the original published post, and replacing the post object breaks this context.

amyevans:
What about adding a filter to `wp_irving_components_query_string` (prior to the query being executed) instead of `wp_irving_components_wp_query`? Then you could run most of the same checks (is the user logged in? do we have the preview param and preview_id param?), and if so, just readd `add_filter( 'the_preview', '_set_preview' );` so that the core logic takes over again, and you don't have to grab the revisions yourself and manually replace the title/content/excerpt. That way the _thumbnail_id preview works too, and it's just less custom code overall.

jameswburke:
@amyevans I took a swing at that, but had issues with it. In order to get it triggered as a revision I had to kill the day, month, year, and name parameters from the query string. If you want to pair on this I'd be happy to take another look.

amyevans:
Yeah let's take a look, I took a quick stab at it earlier and it was working, but maybe there's some nuance or use cases I wasn't properly considering

inc/class-previews.php
```diff
@@ -0,0 +1,185 @@
...
+	 * Hook into the component API request and change the draft preview to be
+	 * `publish`, mimicking a public post.
+	 *
+	 * @param \WP_REST_Request $request Request object.
+	 */
+	public function enabled_public_previews( \WP_REST_Request $request ) {
+
+		// Require `preview` to be true, and `p` to be an integer.
+		if (
+			false === wp_validate_boolean( $request->get_param( 'preview' ) )
```
juliobranha, line 93: Didn't know that one :) I learned something new :)

inc/class-previews.php
```diff
@@ -76,6 +77,7 @@ function() {
 	 */
 	public function modify_query_vars( array $vars ): array {
 		$vars[] = 'preview_id';
+		// $vars[] = '_thumbnail_id';
```
jameswburke:
```suggestion
```

PR #134, opened 2020-04-17
Convert to Alley Coding Standards

joemcgill:
This converts the WP_Irving plugin to use Alley PHPCS sniffing rules.
See: https://github.com/alleyinteractive/Alley-Coding-Standards


joemcgill:
Need to fix up the errors, but wanted to play with using this as way to dogfood and improve on our internal standards.

PR #136, opened 2020-04-27
Add initial template loading logic to components endpoint

joemcgill:
This adds the ability to automatically load templates in a theme based on the default WordPress template hierarchy.

By default, this will look in the active theme for a `templates` directory and will load either a `php`, `json`, or `html` file (in that order) based on the current `WP_Query` object. 

For example, if the path is to a generic archive page, this will try to load `/templates/archive.php`, `/templates/archive.json`, `/templates/index.php`, `/templates/index.json`, etc. The path to the template location is filterable using the new `wp_irving_template_path` hook.

If the context is `site`, this will also attempt to load a `defaults.php` or `defaults.json` template.

To test this, add a templates directory to your active theme and add [the example files from this gist](https://gist.github.com/joemcgill/558cd5af0b7c2edc174bc1d5ef9df783).

Hitting the Irving endpoint with a path for a specific post or page should return the `single.json` data, while hitting any other path should show the `index.json` data. Include `context=site` in your REST request to see the response hydrated with data from the defaults.json.

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -169,8 +169,6 @@ class Components_Endpoint extends Endpoint {
 		// Force trailing slashes on paths.
 		$this->force_trailing_slashes();
 
-		$this->data = load_template( $this->query, $this->data );
-
 		/**
 		 * Modify the output of the components route.
 		 *
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -7,6 +7,8 @@
 
 namespace WP_Irving\REST_API;
 
+use WP_Irving\REST_API\Endpoint;
+
 /**
  * Components endpoint.
  */
@@ -166,6 +168,8 @@ class Components_Endpoint extends Endpoint {
 
 		// Force trailing slashes on paths.
 		$this->force_trailing_slashes();
+
+		$this->data = load_template( $this->query, $this->data );
 
 		/**
 		 * Modify the output of the components route.
--- a/inc/templates.php
+++ b/inc/templates.php
@@ -17,8 +17,8 @@ add_filter( 'wp_irving_components_route', __NAMESPACE__ . '\\load_template', 10,
  *
  * Based on wp-includes/template-loader.php.
  *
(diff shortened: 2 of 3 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -7,6 +7,8 @@
 
 namespace WP_Irving\REST_API;
 
+use WP_Irving\REST_API\Endpoint;
```
joemcgill, line 10: Can probably remove. I was getting a PHPCS error in my editor before adding.

inc/endpoints/class-components-endpoint.php
```diff
@@ -359,7 +361,9 @@ public function build_query() {
 		// Execute query.
 		$wp_query = new \WP_Query( $query );
 
-		if ( empty( $wp_query->posts ) && ! $wp_query->is_search() ) {
+		if ( '/' === $this->path ) {
```
joemcgill, line 364: This is not technically correct, because if a site is set to have another page as the page for posts, _that_ page should return true `is_home()` and the root path should not. Instead, the root path should return `is_front_page()`. 

This is tricky because there is some logic in `WP_Query::parse_query()` which sets the `is_home` flag based on a bunch of conditionals—one of which is (annoyingly) that this isn't a REST request. [See the source](https://github.com/WordPress/WordPress/blob/5.3/wp-includes/class-wp-query.php#L960-L962).

Afterward, this flag is used in the logic for several other conditionals, so we either need to manually fix that check when we build the query and then parse the query again, or we need to fork all of the parse query logic that depends on the `is_home` flag. Either way, I think that's outside the scope of this PR so I'm going to log a new issue for it and we can fix in a separate issue.

jameswburke: Sounds good - yeah, this was solely meant to be a quick hotfix for the sake of testing other functionality. I agree it needs more discussion and consideration before a full solution can be developed.

joemcgill: Opened #139 for this.

jameswburke: Do you think this is a fine temporary solution? Or should we remove the "fix" from this PR? I'm okay either way.

PR #140, opened 2020-04-29
Fix deprecated warning for hook.

### LINKED ISSUE #138: Fix deprecated tag_list_actions warning.
Since WP 5.4, a deprecation notice is being shown on taxonomy list tables when the `tag_list_actions` hook is in use. See: https://core.trac.wordpress.org/ticket/49808

In the mean time, we need to loop through all taxonomies and hook individually to `{$taxonomy}_row_actions` instead.

joemcgill:
This loops through all available taxonomies and adds the appropriate `${taxonomy}_row_actions` callback now that the `tag_row_actions` hook is throwing a deprecation warning.

Resolves #138.

### DIFF
--- a/inc/endpoints/class-components-endpoint.php
+++ b/inc/endpoints/class-components-endpoint.php
@@ -88,19 +88,12 @@ class Components_Endpoint extends Endpoint {
 		parent::__construct();
 
 		add_action( 'init', [ $this, 'set_cap' ] );
+		add_action( 'init', [ $this, 'hook_taxonomy_row_actions' ] );
 		add_filter( 'rest_url', [ $this, 'fix_rest_url' ] );
 		add_filter( 'query_vars', [ $this, 'modify_query_vars' ] );
 		add_filter( 'post_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
 		add_filter( 'page_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
 		add_filter( 'admin_bar_menu', [ $this, 'add_api_link_to_admin_bar' ], 999 );
-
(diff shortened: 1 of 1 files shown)


inc/endpoints/class-components-endpoint.php
```diff
@@ -92,8 +92,15 @@ public function __construct() {
 		add_filter( 'query_vars', [ $this, 'modify_query_vars' ] );
 		add_filter( 'post_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
 		add_filter( 'page_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
-		add_filter( 'tag_row_actions', [ $this, 'add_api_link_to_terms' ], 10, 2 );
 		add_filter( 'admin_bar_menu', [ $this, 'add_api_link_to_admin_bar' ], 999 );
+
+		// Set up taxonomy row actions.
+		// See: https://core.trac.wordpress.org/ticket/49808.
+		add_action( 'init', function () {
```
kevinfodness, line 99: I would recommend that the callback be moved to a named function within the class, similar to `[ $this, 'add_api_link_to_terms' ]`. That way, the function can be called directly in unit tests and is easier to debug.

joemcgill: I normally I agree with making hook callbacks named functions. In this case all that function is doing is adding another callback to the correct hooks, so there's not really anything to test other than that the `add_api_link_to_terms` callbacks are being hooked.

I don't have a strong preference though, so I'll defer to your suggestion since I was on the fence about it.

joemcgill:
BTW, this looks like it might get reverted in a future WP minor release.

PR #145, opened 2020-05-04
Component registry and template parser

jameswburke:
* Creates a component registry and endpoint.
* Template parsing proof of concepts.
** Include template parts.
** Data provider.
** Config and component callbacks.


inc/templates.php
```diff
@@ -274,3 +279,159 @@ function convert_blocks_to_components( array $blocks ): array {
...
+
+	foreach ( $components as $index => &$component ) {
+
+		// This allows us to set text nodes.
+		if ( is_string( $component ) ) {
+			continue;
+		}
+
+		// Ensure we have all the right field and types.
+		$component = ensure_component_fields_exist( $component );
```
joemcgill, line 299: I wonder if instantiating a `WP_Component` object here would allow us to avoid needing to juggle validation by handling it inside the component object?

inc/templates.php
```diff
@@ -274,3 +279,159 @@ function convert_blocks_to_components( array $blocks ): array {
...
 	return $components;
 }
+
+/**
+ * Recursively iterate on a component tree.
+ *
+ * @param array $components Array of components.
+ * @return array
+ */
+function traverse_components( array $components ): array {
```
joemcgill, line 289: I think we can iterate on the hydration logic here, but this looks like a good start. I think it would be helpful for us to build some basic unit tests out of a few basic use cases to ensure that this is doing what we expect and that we don't introduce regressions as we mature this functionality.

Generally speaking, it would be nice if we could get this down to a process wherein hydration is done in a few contained steps:

1. Resolve any template partials to construct a template tree
2. Recursively loop through template tree
3. Instantiate `WP_Component` objects for each component found, passing configuration from the template to the constructor of the object in order to set up the right data.
4. Return hydrated data objects using the callback function from each registered component in the tree.

inc/components/class-registry.php
```diff
@@ -0,0 +1,115 @@
...
+	 * @var null|self
+	 */
+	protected static $instance;
+
+	/**
+	 * Get class instance
+	 *
+	 * @return self
+	 */
+	public static function instance() {
```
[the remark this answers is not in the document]
jameswburke: Updated the class, let me know your thoughts.

inc/components/class-registry.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param string $name Component name.
+	 * @return array
+	 */
+	public function get_registered_component( string $name ) {
+		return $this->components[ $name ] ?? null;
+	}
+}
+
+/**
```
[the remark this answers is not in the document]
jameswburke: Removed.

inc/components/registration/archive-components.php
```diff
@@ -0,0 +1,28 @@
...
+ * @param array $component Component
+ * @return string
+ */
+register_component(
+	'archive/title',
+	[
+		'callback' => function( $component ) {
+			return get_the_archive_title();
+		},
+		'data_provider' => [
```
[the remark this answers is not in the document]
jameswburke: We have a ticket for this, I'd like to merge this proof of concept and iterate on it.

inc/templates.php
jameswburke: Agreed, we have a ticket for this.

inc/templates.php
jameswburke: I agree, and I'd like to merge what we have for now and revisit to improve the base logic.

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
+<?php
+/**
+ * Component registry.
+ *
+ * @package WP_Irving
+ */
+
+namespace WP_Irving\Components;
+
+/**
```
joemcgill, line 10: The `class-registry.php` file should only define the class. Any additional helper functions should go in a `registry.php` file or a generic `inc/components/functions.php` file.

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
...
+		'wp_irving_component_registry_directories',
+		[
+			'wp_irving' => WP_IRVING_PATH . '/inc/components/', // Load components from WP Irving.
+			'theme'     => get_stylesheet_directory() . '/components/', // Load components from the activated theme.
+		]
+	);
+
+	get_registry()->load_components( $directories );
+}
+add_filter( 'init', __NAMESPACE__ . '\auto_register_components' );
```
joemcgill, line 46: Not a blocker, but I wonder if we should add our own `irving_init` action hook so that these types of things only get loaded if Irving is doing it's thing?

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	/**
+	 * Get a single registered component by name.
+	 *
+	 * @param string $name Component name.
+	 * @return array
+	 */
+	public function get_registered_component( string $name ): ?array {
+		return $this->components[ $name ] ?? null;
```
joemcgill, line 76: The empty return value needs to be `[]` or the docblock needs to be updated.

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @param array  $args Component args.
+	 */
+	public function register_component( string $name, array $args = [] ) {
+		$this->components[ $name ] = $args;
+	}
+
+	/**
+	 * Register a component using a json config.
+	 *
+	 * @param string $json_file JSON config file.
```
joemcgill, line 92: Is this a full path?

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
...
+		$config = json_decode( $config, true );
+
+		// Validate config loaded and `name` is available.
+		if ( is_null( $config ) || ! isset( $config['name'] ) ) {
+			return false;
+		}
+
+		$this->register_component(
+			$config['name'],
+			array_merge_recursive( $config, $args )
```
joemcgill, line 118: When would we need to register a component from config AND overload the config with additional arguments?

inc/class-registry.php
```diff
@@ -0,0 +1,152 @@
...
+		return true;
+	}
+
+	/**
+	 * Loop through some directories importing components and registering them.
+	 *
+	 * @param array $directories Directoriess to recursively loop through and
+	 *                           load from.
+	 */
+	public function load_components( array $directories = [] ) {
```
joemcgill, line 130: This is very cool 💖

inc/endpoints/class-components-registry-endpoint.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Components Registry endpoint.
+ */
+class Components_Registry_Endpoint extends Endpoint {
+
+	/**
+	 * Register the REST API routes.
+	 */
+	public function register_rest_routes() {
```
joemcgill, line 20: 👨‍🍳💋

inc/endpoints/class-components-registry-endpoint.php
```diff
@@ -0,0 +1,34 @@
...
+				'methods'  => \WP_REST_Server::READABLE,
+				'callback' => function() {
+					return \WP_Irving\Components\get_registry()->get_registered_components();
+				},
+			]
+		);
+	}
+}
+
+new Components_Registry_Endpoint();
```
joemcgill, line 34: Don't instantiate the class when the file is included. This should happen somewhere else. I'd like for us to follow the [PSR-1 standard](php-fig.org/psr/psr-1/) here:

```
Files SHOULD either declare symbols (classes, functions, constants, etc.) or cause side-effects (e.g. generate output, change .ini settings, etc.) but SHOULD NOT do both.
```

inc/templates.php
```diff
@@ -191,11 +195,12 @@ function filter_template_loader() {
...
  * Return a located template file.
  *
- * @param array $templates A list of possible template files to load.
+ * @param array  $templates A list of possible template files to load.
+ * @param string $folder    Subfolder to look in.
  * @return string The path to the found template.
  */
-function locate_template( array $templates ): string {
-	$template_path = STYLESHEETPATH . '/templates/';
+function locate_template( array $templates, $folder = '/templates/' ): string {
```
joemcgill, line 202: Rather than passing `$folder` I think we should have separate functions for loading templates and loading template partials so we can handle them separately. A `load_template_part()` function could simply look in the `template-parts` directory for the existence of the specific file being loaded and return the components, rather than resolving the existence of a file from an array of possibilities.

inc/templates.php
```diff
@@ -274,3 +279,212 @@ function convert_blocks_to_components( array $blocks ): array {
...
+ * @param array $component Component.
+ * @return array
+ */
+function ensure_component_fields_exist( $component ) {
+
+	if ( is_string( $component ) ) {
+		return $component;
+	}
+
+	return [
```
joemcgill, line 331: This is fine, but would be a good spot to make use of `wp_parse_args()` for merging default config values with those present in the component.

inc/templates.php
```diff
@@ -274,3 +279,212 @@ function convert_blocks_to_components( array $blocks ): array {
...
+ *   "name": "example",
+ *   "config": {
+ *     "href": "{{post/permalink}}"
+ *   }
+ * }
+ *
+ * @param array $component Component.
+ * @return array
+ */
+function handle_component_config_callbacks( $component ): array {
```
joemcgill, line 441: This is super helpful to see how you were thinking about this. I'd still prefer to find a way of handling this without introducing non-JSON-y concepts if possible but am ok keeping this as experimental for now while we're playing around unless you're ok with pulling this out and making this a follow-up PR as we're messing with a context API. Your call.

inc/class-registry.php
jameswburke: `?array` indicates it can be an `array` or `null` right?

inc/class-registry.php
jameswburke: The recursive may be overkill, but one example is the callback being merged into json config.

inc/templates.php
jameswburke: There's two reasons I did it this way,

1. So we only have the component fields, and other stuff gets filtered out (and I acknowledge there are a few ways to do this).
1. Typecasting

inc/templates.php
jameswburke: I'd like to include it for now and trash it after the context stuff is setup, it's useful as a temporary solution so we're not blocked on experiments.

inc/class-registry.php
jameswburke: Components could have other functionality that's required outside of an endpoint context, so I'd like to handle this separately from this PR with some additional consideration/discussion.

inc/class-registry.php
jameswburke: Updated the param to clarify what it is.

inc/templates.php
jameswburke: Added another function to specifically handle template parts as you suggest. It doesn't feel very dry, but let me know your thoughts.

jameswburke:
@joemcgill I've pushed some updates for this based on your comments.

inc/components/registration/archive-components.php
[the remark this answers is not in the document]
joemcgill: Yup. That was my intent. Was just trying to get me thoughts down in more detail here.

PR #150, opened 2020-05-13
Component class

jameswburke:
* Implements a new version of the [base component class from WP Components](https://github.com/alleyinteractive/wp-components/blob/master/inc/wp-components/classes/class-component.php).
* Template parser uses component class for validation.

More unit tests and additional functionality coming later.

### DIFF
--- a/inc/class-component.php
+++ b/inc/class-component.php
@@ -258,7 +258,7 @@ class Component implements \JsonSerializable {
 	}
 
 	/**
-	 * Set all children
+	 * Set all children.
 	 *
 	 * @param array $children Children.
 	 * @return self
@@ -278,7 +278,7 @@ class Component implements \JsonSerializable {
 		return $this->set_children(
 			array_merge(
 				$this->sanitize_children( $children ),
-				$this->children
+				$this->get_children()
 			)
 		);
 	}
@@ -292,7 +292,7 @@ class Component implements \JsonSerializable {
 	public function append_children( array $children ): self {
 		return $this->set_children(
 			array_merge(
-				$this->children,
+				$this->get_children(),
 				$this->sanitize_children( $children )
 			)
 		);
@@ -315,12 +315,7 @@ class Component implements \JsonSerializable {
 	 * @return self
 	 */
 	public function prepend_child( $child ): self {
-		return $this->set_children(
-			array_merge(
-				$this->sanitize_children( [ $child ] ),
-				$this->children
-			)
-		);
+		return $this->prepend_children( [ $child ] );
 	}
 
 	/**
@@ -330,12 +325,7 @@ class Component implements \JsonSerializable {
 	 * @return self
 	 */
 	public function append_child( $child ): self {
-		return $this->set_children(
-			array_merge(
-				$this->children,
-				$this->sanitize_children( [ $child ] )
-			)
-		);
+		return $this->append_children( [ $child ] );
 	}
 
 	/**
@@ -361,8 +351,8 @@ class Component implements \JsonSerializable {
 	/**
 	 * Set the component theme.
 	 *
-	 * @param string       $theme Theme name.
-	 * @param bool|boolean $force Ignore the theme options.
+	 * @param string $theme Theme name.
+	 * @param bool   $force Optional. Ignore the theme options. Default false.
 	 * @return self
 	 */
 	public function set_theme( string $theme, bool $force = false ): self {
@@ -451,7 +441,7 @@ class Component implements \JsonSerializable {
 					function( $theme ) {
 						$theme = (string) $theme;
 						$theme = trim( $theme );
-						// @todo Should this be camelcased?
+						$theme = self::camel_case( $theme );
 						return $theme;
 					},
 					$this->theme_options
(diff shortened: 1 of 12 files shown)


inc/class-component.php
```diff
@@ -0,0 +1,618 @@
...
+	 * @return array Updated array with camel-cased keys.
+	 */
+	public function camel_case_keys( $array ): array {
+
+		// Setup for recursion.
+		$camel_case_array = [];
+
+		// Loop through each key.
+		foreach ( $array as $key => $value ) {
+			// Only return keys that are white-listed. Leave $whitelist empty
```
joemcgill, line 547: Let's take this opportunity to remove non-inclusive "white-list/black-list" terminology in favor of "allowed/blocked" terms. It doesn't seem like there even is a `whitelist` property for this class at the moment, so this will break.

inc/class-component.php
```diff
@@ -0,0 +1,618 @@
...
+			// Reassemble key.
+			$new_key = implode( '', $words );
+
+			// Lowercase the first character.
+			$new_key[0] = strtolower( $new_key[0] );
+
+			if (
+				! is_array( $value )
+				// Don't recursively camelCase if this key is in the $preserve_inner_keys property.
+				|| ( ! empty( $this->preserve_inner_keys ) && in_array( $key, $this->preserve_inner_keys, true ) )
```
joemcgill, line 577: There isn't a `preserve_inner_keys` property on this class.

inc/templates.php
```diff
@@ -478,18 +480,14 @@ function handle_data_provider( $component ) {
  * @param array $component Component.
  * @return array
  */
-function handle_component_config_callbacks( $component ): array {
+function handle_component_config_callbacks( $component ): Component {
```
joemcgill, line 483: We should be able to deprecate this in favor of the coming context API.

tests/test-component.php
```diff
@@ -0,0 +1,86 @@
...
+class Component_Tests extends WP_UnitTestCase {
+
+	/**
+	 * Associative array of components to use for testing.
+	 *
+	 * @var array
+	 */
+	public static $components = [];
+
+	public function setUp() {
```
joemcgill, line 22: `setUp()` will rerun the `create_test_components()` method before each test method. To set these up once and share them across methods, `wpSetUpBeforeClass()` would be a better choice.

inc/templates.php
```diff
@@ -8,6 +8,7 @@
 namespace WP_Irving\Templates;
 
 use WP_Query;
+use \WP_Irving\Component;
```
joemcgill, line 11: Don't need the initial `\` here.

inc/class-component.php
```diff
@@ -0,0 +1,618 @@
...
+	}
+
+	/**
+	 * Set a config value by key, or the entire config array.
+	 *
+	 * @param array|string $config_array_or_key Config array or key.
+	 * @param mixed        $value               Config key value.
+	 * @return self
+	 */
+	public function set_config( $config_array_or_key, $value = null ): self {
```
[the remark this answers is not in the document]
jameswburke: Yeah, and `merge_config()` could be used for that.

joemcgill:
Looks like the PHPUnit tests are failing because of we're not including the legacy redirector plugin when these tests are run. My opinion would be to check for whether that plugin is loaded and if not mark that test class as skipped during the `setUp()` method of that test class. See: https://phpunit.readthedocs.io/en/9.0/incomplete-and-skipped-tests.html

inc/class-component.php
```diff
@@ -0,0 +1,618 @@
...
+	 * @return function
+	 */
+	public function callback( $callable, ...$args ) {
+		return call_user_func_array( $callable, array_merge( [ &$this ], $args ) );
+	}
+
+	/**
+	 * Execute the callback method in this component's registration.
+	 */
+	public function execute_registry_callback(): self {
```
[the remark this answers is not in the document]
jameswburke: I don't fully understand this, so I'm going to rollback the change and we can discuss in a future PR.

inc/templates.php
jameswburke: How so? I don't see how the context api removes the need for a render callback?

inc/class-component.php
```diff
@@ -0,0 +1,618 @@
...
+
+	/**
+	 * Prepend children.
+	 *
+	 * @param array $children Children.
+	 * @return self
+	 */
+	public function prepend_children( array $children ): self {
+		return $this->set_children(
+			array_merge(
```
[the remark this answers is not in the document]
jameswburke: From my testing, `array_merge` is the result I'm looking for,

Example,
```<?php
$children = [];

$additional_children = [ 'component_one', 'component_two' ];

$merged_children = array_merge(
    $children,
    $additional_children
);

array_push( $children, $additional_children );

echo 'Push:</br>';
print_r( $children );

echo '</br></br>Merge:</br>';
print_r( $merged_children );

?>```

```Push:
Array ( [0] => Array ( [0] => component_one [1] => component_two ) )

Merge:
Array ( [0] => component_one [1] => component_two )```

File: inc/templates.php (hunk 3 above)
Comment (joemcgill) [reply inc/templates.php:483]: Maybe I'm misunderstanding the use of this function. I thought the responsibility of this function was to convert the mustache style config values to computed values.

File: inc/class-component.php (hunk 8 above)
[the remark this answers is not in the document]
Comment (joemcgill) [reply inc/class-component.php:279]: Ah, didn't see that you were adding multiple values to the end of the array. All good 👍

File: inc/templates.php (hunk 3 above)
Comment (jameswburke) [reply inc/templates.php:482]: Right, I'm getting my wires crossed. And yes, we'll be deprecating soon. Going to leave as-is for now though.

**Comment by jameswburke** [comment]
@joemcgill I think I've got everything you mentioned + fixed the redirector unit tests.

File: tests/test-component.php [hunk 9]
```diff
@@ -83,4 +82,71 @@ public function test_name() {
...
+	 * prepend_child()
+	 * append_child()
+	 * sanitize_children()
+	 */
+	public function test_children() {
+
+		$parent = $this->get_component( 'children-test-001' );
+		$child  = $this->get_component( 'children-test-002' );
+
+		// get_children()
```
joemcgill, line 104: Good to note these as todos. When implementing, each test should be handled in a separate test method rather than testing all of these within a single method.

tests/test-component.php
```diff
@@ -83,4 +82,71 @@ public function test_name() {
...
+			'upperCase'         => '',
+			'recursiveTest'     => [
+				'helloWorld2' => '',
+			],
+		];
+
+		$camel_case_keys = ( new Component() )->camel_case_keys( $snake_case_keys );
+
+		$this->assertEquals( $expected_camel_case_keys, $camel_case_keys, 'Top level keys don\'t match.' );
+		$this->assertEquals( $expected_camel_case_keys['recursiveTest'], $camel_case_keys['recursiveTest'], 'Nested keys don\'t match.' );
```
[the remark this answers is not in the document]
jameswburke: It didn't seem to work for me, but I'll give it another go.

PR #152, opened 2020-05-15
Continue writing tests

jameswburke:
This is a work in progress. Eventually the entire class will have testing. Incremental PR for ease of reviewing.

* Refactors the theme functionality.
* Adds more testing.


inc/class-component.php
```diff
@@ -378,80 +378,119 @@ public function get_theme_options(): array {
...
 	 * @return self
 	 */
-	public function sanitize_theme_options(): self {
-
-		$this->theme_options = array_unique(
-			array_filter(
-				array_map(
-					function( $theme ) {
-						$theme = (string) $theme;
-						$theme = trim( $theme );
```
davisshaver, line 443: Does the trimming and casting still need to occur?

inc/class-component.php
```diff
@@ -378,80 +378,119 @@ public function get_theme_options(): array {
...
 		);
 
 		return $this;
 	}
 
+	/**
+	 * Remove theme option.
+	 *
+	 * @param string $theme_option Theme name.
+	 * @return [type]               [description]
```
davisshaver, line 469: Missing docs?

inc/class-component.php
```diff
@@ -585,14 +624,24 @@ public function jsonSerialize(): array {
...
+		// Camel case theme options.
+		$this->set_theme_options( $this->camel_case_keys( $this->get_theme_options() ) );
+		$this->set_theme( self::camel_case( $this->get_theme() ) );
+
 		// Add the theme name to the config as Irving core expects.
 		$this->set_config( 'theme_name', $this->get_theme() );
 		$this->set_config( 'theme_options', $this->get_theme_options() );
 
+		// Sanitize children.
+		$this->set_children( self::reset_array( $this->get_children() ) );
```
davisshaver, line 639: is the interior `self::reset_array()` needed here as that seems to be what `$this->set_children()` does?

tests/test-component.php
```diff
@@ -63,90 +66,574 @@ public static function create_test_components() {
 	 */
 	public function get_component( $name ) {
 		if ( isset( self::$components[ $name ] ) ) {
-			return self::$components[ $name ];
+			return clone self::$components[ $name ];
```
davisshaver, line 69: might be worth a note here, I guess this lets each test run on a fresh component?

tests/test-component.php
```diff
@@ -63,90 +66,574 @@ public static function create_test_components() {
...
+		// Test adding two.
+		$component->add_theme_options( [ 'extra-large', 'double-extra-large' ] );
+		$this->assertEquals(
+			[
+				'default',
+				'primary',
+				'secondary',
+				'large',
+				'extra-large',
+				'double-extra-large',
```
davisshaver, line 397: 🍟

tests/test-component.php
```diff
@@ -63,90 +66,574 @@ public static function create_test_components() {
...
+				'fooBar' => '',
 			],
 		];
 
+		// Run the method.
 		$camel_case_keys = ( new Component() )->camel_case_keys( $snake_case_keys );
 
 		$this->assertEquals( $expected_camel_case_keys, $camel_case_keys, 'Top level keys don\'t match.' );
-		$this->assertEquals( $expected_camel_case_keys['recursiveTest'], $camel_case_keys['recursiveTest'], 'Nested keys don\'t match.' );
+		// $this->assertEquals( $expected_camel_case_keys['recursiveTest'], $camel_case_keys['recursiveTest'], 'Nested keys don\'t match.' );
```
davisshaver, line 603: maybe add comment here for commented out code?

inc/class-component.php
jameswburke: I decided to sanitize on output rather than what gets stored. It was becoming clunky otherwise, and may lead to unexpected dev experience.

inc/class-component.php
jameswburke: I have it as a kind of "catch all in case something suspect happens", but I think it can be removed for now and addressed later if needed. Thanks.

tests/test-component.php
jameswburke: Yeah. It's cheaper than reloading the components somehow.

PR #153, opened 2020-05-15
More tests

jameswburke:
* More test coverage for component class.
* Method to hydrate children.


inc/class-component.php
```diff
@@ -539,7 +570,7 @@ public function set_context_consumer( array $context_consumer ): self {
 	 * @param mixed    ...$args  Optional args to pass to the callback.
 	 * @return function
```
renatonascalves, line 571:
```suggestion
	 * @return Component
```

inc/class-component.php
```diff
@@ -328,14 +328,45 @@ public function append_child( $child ): self {
...
+			// Use this recursively.
+			if ( $recursive ) {
+				$child->hydrate_children();
+			}
+		}
+
+		$this->set_children( $children );
+
+		return $this;
+	}
```
renatonascalves, line 360:
```suggestion
		return $this->set_children( $children );
	}
```

PR #157, opened 2020-05-19
IRV-418: Template updates and Helmet implementation

jameswburke:
* Adds <Helmet> tags when WP Irving templating is use.
* Filters the Helmet tags for Yoast.
* Small tweaks to templates.


inc/templates.php
```diff
@@ -25,6 +25,10 @@
  * @return array A hydrated data object.
  */
 function load_template( array $data, WP_Query $query, string $context ): array {
+
+	// Manage helmet when using template functionality.
+	add_filter( 'wp_irving_components_route', __NAMESPACE__ . '\\setup_helmet', 11, 5 );
```
joemcgill, line 30: If we're always setting up header info during template loading, we should just call `setup_helmet()` directly when needed, rather than adding another filter hook during a callback, which makes testing more difficult.

inc/templates.php
```diff
@@ -363,14 +448,21 @@ function hydrate_components( array $components ) {
 		$component->provide_context( get_template_context() );
 
 		// Recursively hydrate children.
-		if ( ! empty( $component->get_children() ) ) {
+		if ( ! is_string( $component ) && ! empty( $component->get_children() ) ) {
```
joemcgill, line 451: If we're converting a string to an `irving/text` component in `setup_component()` then I'm not sure when this would ever be a string.

inc/templates.php
```diff
@@ -363,14 +448,21 @@ function hydrate_components( array $components ) {
...
 		}
 
 		// Reset context to where it was before hydration.
 		get_template_context()->reset();
 
-		$hydrated[] = $component->jsonSerialize();
+		// Convert text nodes to actual text notes.
+		// @todo there's _definitely_ a better way to do this, but certain
+		// Irving core functionality won't work without this hack.
+		if ( 'irving/text' === $component->get_name() ) {
```
joemcgill, line 461: We might be able to handle this the same way we were talking about handling silent components, but cool to leave special handling for now.

inc/templates.php
```diff
@@ -391,7 +483,7 @@ function setup_component( $component ) {
 	// Convert strings to text components.
 	if ( is_string( $component ) ) {
 		$component = [
-			'name' => 'text',
+			'name' => 'irving/text',
```
joemcgill, line 486: 👍

inc/templates.php
```diff
@@ -483,7 +575,15 @@ function hydrate_template_parts( $component ) {
 		return false;
 	}
 
-	return hydrate_components( prepare_data_from_template( $template ) );
+	$template_data = prepare_data_from_template( $template );
+
+	// Handle template parts with only one component rather than an array of
+	// components.
+	if ( isset( $template_data['name'] ) ) {
```
joemcgill, line 582: `prepare_data_from_template()` should always return an array. If you're seeing a case where that doesn't happen, let's handle it within that function.

inc/templates.php
jameswburke: I was thinking other sites may want to enable this, and/or some sites may want to disable it?

inc/templates.php
jameswburke: It's returning an array, depending if the template part is an array (multiple components) vs. object (single component), then we need to wrap the single array in another array so it hydrates properly, otherwise it includes the template part weird.

I'll try to add a test for this.

inc/templates.php
jameswburke: Remnant from an older PR from before we chatted.

inc/templates.php
joemcgill: Yeah, that's totally reasonable. For functionality that we use by default, but think might need to be disabled, I prefer a pattern where we call it directly during our template loading but add a filter within that function to disable it if not needed instead. So something like:

```
setup_helmet( $data ) {

	$enabled = apply_filters( 'wp_irving_setup_helmet', true, $data );

	// Bail early if disabled.
	if ( ! $enabled ) {
	  return $data
	}

	// helmet logic...
}
```

This has two benefits, it makes the execution flow more predictable (we're not relying on the callback order of hooks) and makes the inclusion of helmet in the template logic more testable because our functions aren't producing side effects.

jameswburke:
@joemcgill pushed some tweaks some new tests.

tests/test-helmet.php
```diff
@@ -0,0 +1,317 @@
...
+		$parsed_html = parse_html( $this->example_markup, [ 'title' ] );
+
+		$this->assertEquals(
+			( new Component( 'irving/helmet' ) )->set_child( $title ),
+			inject_head_tags( $helmet, $parsed_html )
+		);
+	}
+
+	/**
+	 * Test the `get_title_from_helmet()` functin.
```
joemcgill, line 145: Typo: s/functin/function

tests/test-helmet.php
```diff
@@ -0,0 +1,317 @@
...
+<link rel="canonical" href="https://irving.alley.test/" />
+<meta name="googlebot" content="index, follow, max-snippet:-1, max-image-preview:large, max-video-preview:-1" />
+<meta name="bingbot" content="index, follow, max-snippet:-1, max-image-preview:large, max-video-preview:-1" />
+<script type="application/ld+json" class="yoast-schema-graph">{"@context":"https://schema.org","@graph":[{"@type":"WebSite","@id":"https://irving.alley.test/#website","url":"https://irving.alley.test/","name":"Irving Development","description":"Just another WordPress site","potentialAction":[{"@type":"SearchAction","target":"https://irving.alley.test/?s={search_term_string}","query-input":"required name=search_term_string"}],"inLanguage":"en-US"},{"@type":"CollectionPage","@id":"https://irving.alley.test/#webpage","url":"https://irving.alley.test/","name":"Irving Development - Just another WordPress site","isPartOf":{"@id":"https://irving.alley.test/#website"},"description":"Just another WordPress site","inLanguage":"en-US"}]}</script>
+<!-- / Yoast SEO plugin. -->';
+
+	/**
+	 * Tests the automatic insertion and filters for <Helmet>.
+	 */
+	function test_setup_helmet() {
```
joemcgill, line 42: These are good tests, but there's a lot going on in this one test method. Let's try and break these up so that generally there is only one assertion (or type of assertion) being made per test method, or reuse the same test method to iterate over different data using a data provider.

tests/test-helmet.php
```diff
@@ -0,0 +1,317 @@
...
+		$this->assertEquals(
+			( new Component( 'irving/helmet' ) )->set_child( $title ),
+			inject_head_tags( $helmet, $parsed_html )
+		);
+	}
+
+	/**
+	 * Test the `get_title_from_helmet()` functin.
+	 */
+	public function test_get_title_from_helmet() {
```
joemcgill, line 147: This is a good candidate for a data provider. When you test several different iterations of test data it's also helpful to pass an extra message parameter to the assertion that describes which test failed, e.g.:

```
$helmet = new Component( 'irving/helmet' );
$this->assertNull( get_title_from_helmet( $helmet ), 'Title not null in default helmet component.' );
```

tests/test-helmet.php
```diff
@@ -0,0 +1,317 @@
...
+						'children' => [ 'Hello World' ]
+					]
+				)
+			);
+	}
+
+	/**
+	 * Test the `create_or_update_title` function.
+	 */
+	public function test_create_or_update_title() {
```
joemcgill, line 213: This could also be broken up into a data provider or perhaps be broken into distinct methods to test creation and updating paths.

tests/test-templates.php
```diff
@@ -281,4 +281,38 @@ function test_templates_hydrate_partials() {
...
+					'themeOptions' => [ 'default' ],
+				],
+				'children' => [],
+			],
+		];
+
+		$this->assertEquals(
+			$expected,
+			Templates\hydrate_components( $template ),
+			'Could not get context in non-registered components.'
```
joemcgill, line 315: I think this message needs to be updated.

jameswburke:
@joemcgill I refactored many of the tests into their own methods, created some additional helpers/data providers,  and added messages to the asserts.

PR #158, opened 2020-05-20
Add origin to cache request

ostowe:
Question: Seems like this request should already have `site_url()` as its origin, am I missing something?


ostowe:
Hh interesting, cool. Well, this works for now, but perhaps there's some refinement we can make in the future.

PR #163, opened 2020-05-26
More tweaking

jameswburke:
A few small tweaks for errors/issues I hit during development.

### DIFF
--- a/inc/class-component.php
+++ b/inc/class-component.php
@@ -212,8 +212,8 @@ class Component implements JsonSerializable {
 	public function set_config( $config_array_or_key, $value = null ): self {
 
 		// Set the entire config.
-		if ( is_array( $config_array_or_key ) ) {
-			$this->config = $config_array_or_key;
+		if ( is_array( $config_array_or_key ) || is_object( $config_array_or_key ) ) {
+			$this->config = (array) $config_array_or_key;
 			return $this;
 		}
 
--- a/inc/templates.php
+++ b/inc/templates.php
@@ -458,21 +458,48 @@ function parse_config_from_registry( array $component ) {
 	$parsed_config = [];
 
 	$type_callbacks = [
-		'array'  => 'is_array',
-		'bool'   => 'is_bool',
-		'int'    => 'is_int',
-		'number' => 'is_numeric',
-		'string' => 'is_string',
-		'text'   => 'is_string',
+		'array'   => 'is_array',
+		'bool'    => 'is_bool',
+		'int'     => 'is_int',
+		'integer' => 'is_int',
+		'number'  => 'is_numeric',
+		'string'  => 'is_string',
+		'text'    => 'is_string',
 	];
 
 	// Loop through registered config.
 	foreach ( ( $registered['config'] ?? [] ) as $key => $atts ) {
 
+		// If the config's type is not registered, throw a fatal.
+		if ( ! isset( $type_callbacks[ $atts['type'] ] ) ) {
+			wp_die(
+				sprintf(
+					// Translators: %1$s - Config key name, %2$s - Incorrect type, %3$s - Component namme, %4$s allowed types as string.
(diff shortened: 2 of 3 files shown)


inc/class-component.php
```diff
@@ -212,8 +212,8 @@ public function get_config_by_key( string $key ) {
 	public function set_config( $config_array_or_key, $value = null ): self {
 
 		// Set the entire config.
-		if ( is_array( $config_array_or_key ) ) {
-			$this->config = $config_array_or_key;
+		if ( is_array( $config_array_or_key ) || is_object( $config_array_or_key ) ) {
+			$this->config = (array) $config_array_or_key;
```
jameswburke, line 216: If the hydration process has already executed, then this could be an object already. Ideally we wouldn't hydrate more than once, but it's not quite that simple. I think we need to discuss that process a bit and discuss obstacles like this one.

inc/templates.php
```diff
@@ -466,10 +466,14 @@ function parse_config_from_registry( array $component ) {
...
-	foreach ( $registered['config'] ?? [] as $key => $atts ) {
+	// Loop through registered config.
+	foreach ( ( $registered['config'] ?? [] ) as $key => $atts ) {
+
+		// This value has been set and a sanitize callback exists.
 		if (
-			isset( $component['config'][ $key ] ) &&
-			( call_user_func( $type_callbacks[ $atts['type'] ], $component['config'][ $key ] ) )
+			isset( $component['config'][ $key ] )
+			&& isset( $type_callbacks[ $atts['type'] ] )
```
joemcgill, line 475: Not sure what you are trying to solve for here. Are you trying to make `type` an optional value when registering config values, or did you run across an unsupported type that broke here? If the latter, I'd prefer that we add additional supported types/callbacks above rather than skipping any that we can't validate.

inc/templates.php
```diff
@@ -401,7 +401,7 @@ function hydrate_components( array $components ) {
 		if ( 'irving/text' === $component->get_name() ) {
 			$hydrated[] = $component->get_config( 'content' );
 		} else {
-			$hydrated[] = $component->jsonSerialize();
+			$hydrated[] = $component;
```
joemcgill, line 404: I don't understand this change. Is serialization happening elsewhere now?

jameswburke: Serialization will happen automatically when the endpoint is spit out, but converting the components back to arrays before the output is ready is causing issues where I'm trying to modify something, and in some cases it's an array, and in others it's an object. Additionally, when attempting to run the hydration process on something that's already been hydrated, I'm running into issues.

I understand what you're trying to accomplish here, but it's not working well during my experiments, and making this change has cleared things up.

inc/templates.php
jameswburke: I had a typo of `arrays` instead of `array` and it was throwing an error. This ensures that the callback exists before we try to use it.

joemcgill: In that case, seems like we need better error handling instead of working around errors. Otherwise, this seems to have worked as it should have, no?

jameswburke:
@renatonascalves serialization will be triggered when the class instance is type cast to an array, which happens when the data is passed to the rest response. The WP Components class never calls that method directly, for our purposes, it just happens.

inc/templates.php
jameswburke: The only issue here is that it was attempting to call a `$type_callbacks[]` key that didn't exist. This adds a check to ensure it exists.

The actual calling of the method works fine.

jameswburke: Or are you suggesting something like a wp_die printing out an error message suggesting the developer check their "type" key? As I continue to think about it, that makes more sense to me.

jameswburke: Updated in https://github.com/alleyinteractive/wp-irving/pull/163/commits/08b040a7e730a4f74a8237ccb631527cab0f50fb

PR #167, opened 2020-05-29
Begin implementing schema on components

jameswburke:
* Define a schema shape, and set it during the component registration process.
* Introduce filters for `jsonSerialize()`.
* If a config key has a `hidden` property set to true, set it to null upon serialization (so we're exposing sensitive data).
* Unit tests.

### DIFF
--- a/inc/class-component.php
+++ b/inc/class-component.php
@@ -798,8 +798,8 @@ class Component implements JsonSerializable {
 		$this->set_config( 'theme_options', array_keys( $this->camel_case_keys( array_flip( $this->get_theme_options() ) ) ) );
 
 		// Null any config keys where the schema has hidden = true.
-		foreach ( $this->get_schema() as $key => $schmea ) {
-			if ( $schmea['hidden'] ) {
+		foreach ( $this->get_schema() as $key => $schema ) {
+			if ( $schema['hidden'] ) {
 				$this->set_config( $key, null );
 			}
 		}
--- a/inc/class-component.php
+++ b/inc/class-component.php
@@ -105,6 +105,7 @@ class Component implements JsonSerializable {
 			$args,
 			[
 				'config'           => [],
+				'schema'           => [],
 				'children'         => [],
 				'theme'            => 'default',
 				'theme_options'    => [ 'default' ],
@@ -116,6 +117,9 @@ class Component implements JsonSerializable {
 
 		// Set up config.
 		$this->set_config( $args['config'] );
+
+		// Set up schema.
+		$this->set_schema( $args['schema'] );
 
 		// Set up children.
 		$this->set_children( $args['children'] );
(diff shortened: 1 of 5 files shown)


tests/test-templates.php
```diff
@@ -302,38 +302,4 @@ public function test_templates_hydrate_partials() {
 
 		$this->assertEquals( $expected, $hydrated, 'Template partial not hydrated correctly.' );
 	}
-
-	/**
-	 * Tests that context values are passed to non-registered components.
-	 */
-	public function test_templates_with_text_nodes() {
```
jameswburke, line 309: This is now happening as a callback, not as part of the template hydration.

joemcgill:
I'm not sure I understand what you're trying to do here. I assumed that schema for each attribute value would be handled in the `component.json` file when registering a component, and that the shape would look more like:

```
{
	"name": "irving/example",
	"config": {
		"hiddenProp": {
			"default": "",
			"type": "string"
			"hidden": true
		},
}
```

In this example the component gets registered with the expectation that there is a `hiddenProp` config property, which can be set as a `string` via the template or the hydration callback, but that value will not be exposed in the REST response since the property is a `hidden` property.

If I'm understanding correctly, I'm not sure why we need a `schema` property for the whole component.

jameswburke:
@joemcgill, it's always been my intention to add some schema to the component class. I think it's useful to have localized within the class instance's data, rather than reaching out the registry for it (or pass the registered component to the class so it can be referenced).

And while we haven't discussed it yet, I see value in being able to do type validation/defaults upon component serialization as well. All of that functionality _only_ exists within the context of hydration + templating, but could be useful to components elsewhere.

inc/class-component.php
```diff
@@ -740,10 +792,18 @@ public function jsonSerialize(): array {
...
 	 */
 	public function to_array(): array {
+
 		// Add the theme name to the config as Irving core expects.
 		$this->set_config( 'theme_name', self::camel_case( $this->get_theme() ) );
 		$this->set_config( 'theme_options', array_keys( $this->camel_case_keys( array_flip( $this->get_theme_options() ) ) ) );
 
+		// Null any config keys where the schema has hidden = true.
+		foreach ( $this->get_schema() as $key => $schmea ) {
+			if ( $schmea['hidden'] ) {
```
juliobranha, line 802:
```suggestion
		foreach ( $this->get_schema() as $key => $schema ) {
			if ( $schema['hidden'] ) {
```

tests/test-component.php
```diff
@@ -225,6 +225,48 @@ public function test_set_config_by_key() {
...
+
+		$this->assertEquals( $expects, $schema_example->get_schema(), 'Schema did not match what was expected.' );
+	}
+
+	/**
+	 * Tests for the `set_schemaa()` method.
+	 */
+	public function test_set_schema() {
+
+		$new_schmea = [
```
juliobranha, line 251:
```suggestion
		$new_schema = [
```

tests/test-component.php
```diff
@@ -225,6 +225,48 @@ public function test_set_config_by_key() {
...
+			'post' => [
+				'default' => null,
+				'hidden'  => false,
+				'type'    => 'null',
+			],
+		];
+
+		$schema_example = $this->get_component( 'schema' );
+
+		$schema_example->set_schema( $new_schmea );
```
juliobranha, line 265:
```suggestion
		$schema_example->set_schema( $new_schema );
```

inc/class-component.php
```diff
@@ -245,6 +258,32 @@ public function set_config_by_key( string $key, $value ): self {
...
+		return $this->config_schema;
+	}
+
+	/**
+	 * Set schema describing config.
+	 *
+	 * @param array $config_schema Schema array.
+	 * @return self
+	 */
+	public function set_config_schema( array $config_schema ): self {
```
joemcgill, line 276: We may want to make this private and only set schema as an internal implementation detail, but I think it's fine to leave as is for now.

PR #169, opened 2020-06-02
Register components in WP Irving (refactored from example theme)

xmattus:
This is a companion to https://github.com/alleyinteractive/irving/pull/202/

It adds registration and component data for the core Irving components. This is where PHP program logic should live to set up the default hydration patterns for the core library components.

--------------------
* Add a "Registered Components" link to the admin bar.
* Moves candidate component WP code from example Irving theme into this plugin.


inc/components/post-content/component.php
```diff
@@ -0,0 +1,48 @@
...
+			/**
+			 * Taken directly from Gutenberg.
+			 *
+			 * @see https://github.com/WordPress/gutenberg/blob/30cd85aebe14eee995e3162f09d31f4d4786f101/packages/block-library/src/post-content/index.php#L23
+			 */
+			$post_content = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', get_the_content( null, false, $post_id ) ) ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+
+			return $component
+				->set_config( 'content', $post_content )
+				->set_config( 'oembed', true )
```
joemcgill, line 44: We should include oembed in the component.json schema and set the default to `true` rather than setting it here.

inc/components/post-content/component.php
```diff
@@ -0,0 +1,48 @@
...
+				->set_name( 'irving/html' );
```
joemcgill, line 45: This should make use of the `_alias` property rather than changing the name in the hydration callback.

inc/components/post-excerpt/component.php
```diff
@@ -0,0 +1,35 @@
...
+	[
+		'callback' => function( Component $component ): Component {
+
+			// Get the post ID from a context provider, or fallback to the global.
+			$post_id = $component->get_config( 'post_id' );
+			if ( 0 === $post_id ) {
+				$post_id = get_the_ID();
+			}
+
+			return $component->set_config( 'content', html_entity_decode( get_the_excerpt( $post_id ) ) );
```
joemcgill, line 32: We should probably do `apply_filters( 'the_excerpt', get_the_excerpt() );` so this is treated like `the_excerpt()` in WordPress core.

inc/components/post-content/component.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Register the component and callback.
+ */
+get_registry()->register_component_from_config(
+	__DIR__ . '/component',
+	[
+		'callback' => function( Component $component ): Component {
+
+			// Get the post ID from a context provider, or fallback to the global.
```
joemcgill, line 29: Unnecessary since this is set in default context. Same for other post components below 🔢

inc/components/post-featured-media/component.php
```diff
@@ -0,0 +1,75 @@
...
+			if ( empty( $image_url ) ) {
+				return $component;
+			}
+
+			$wrapper = new Component( 'irving/container' );
+			$wrapper->set_config( 'max_width', 'lg' );
+			$wrapper->set_config( 'style', [ 'bottom-margin' => '1rem' ] );
+
+			$wrapper->append_child(
+				( new Component( 'material/card-content' ) )
```
joemcgill, line 45: We shouldn't be using material components in our core library. This may need to be a @todo in another ticket.

inc/components/post-featured-media/component.php
```diff
@@ -0,0 +1,75 @@
...
+									'bottom_margin' => '1rem',
+									'height'        => '450px',
+								]
+							)
+					)
+			);
+
+			$caption = wp_get_attachment_caption( get_post_thumbnail_id( $post_id ) );
+			if ( ! empty( $caption ) ) {
+				$wrapper->append_child(
```
joemcgill, line 62: This should probably just use the default HTML or Text component.

inc/components/text/component.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Register the component and callback.
+ */
+get_registry()->register_component_from_config( __DIR__ . '/component' );
+
+/**
+ * Output the content config value as a text node instead of a component.
+ *
+ * @todo Deterine how text and strings should actually work, and target
```
joemcgill, line 26: Let's not forget this.

jameswburke: It's on the top of my mind. Let's discuss tomorrow.

inc/class-admin.php
```diff
@@ -48,6 +48,7 @@ public function setup() {
 		add_filter( 'post_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
 		add_filter( 'page_row_actions', [ $this, 'add_api_link_to_posts' ], 10, 2 );
 		add_filter( 'admin_bar_menu', [ $this, 'add_api_link_to_admin_bar' ], 999 );
+		add_filter( 'admin_bar_menu', [ $this, 'add_registered_components_link_to_admin_bar' ], 999 );
```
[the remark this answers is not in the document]
jameswburke: Updated to use action.

It's been a long time since this was written, but my guess is I wanted to ensure it was always the last thing, irregardless of what other plugins were doing. Updated to 99.

inc/components/archive-title/component.json
```diff
@@ -0,0 +1,6 @@
+{
+	"name": "irving/archive-title",
+	"alias": "irving/fragment",
+	"category": "Archive",
+	"description": "Get the archive title."
```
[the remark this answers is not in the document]
jameswburke: Updated to use `irving/text`, which will pull the config from `text/component.json`,

inc/components/fragment/component.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
jameswburke: Fragment allows you to pass things through w/o a wrapper, container has a wrapper + other properties. We could refactor later to combine, but I like having the clear distinction for development purposes.

inc/components/post-content/component.php
jameswburke: Updated to alias to `html`, removed the config entirely, and left oembed since it is false by default on `irving/html` (for performance reasons).

inc/components/post-excerpt/component.php
jameswburke: Didn't realize it didn't run through the filters using `get_the_excerpt()`. Good catch. Thanks.

inc/components/post-featured-media/component.php
jameswburke: Created IRV-465 in addition to the @todo.

jameswburke:
@joemcgill I believe that's everything. Let me know how it looks. Thanks.

PR #170, opened 2020-06-04
Feature/default contexts

jameswburke:
Tweak the default context for templates to have the post_id and wp_query.

### DIFF
--- a/inc/templates.php
+++ b/inc/templates.php
@@ -584,22 +584,12 @@ function get_template_context() {
 	if ( empty( $context ) ) {
 		$context = new WP_Irving\Context_Store();
 
-		// Set default context.
-		switch ( true ) {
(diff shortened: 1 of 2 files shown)


inc/templates.php
```diff
@@ -576,13 +575,28 @@ function hydrate_template_parts( $component ) {
...
 function get_template_context() {
+	global $wp_query;
 	static $context;
 
 	if ( empty( $context ) ) {
 		$context = new WP_Irving\Context_Store();
 
 		// Set default context.
-		$context->set( [ 'irving/post' => get_the_ID() ] );
+		switch ( true ) {
```
jameswburke, line 585: Update to have,

- irving/wp
- irving/wp_query
- irving/post

PR #176, opened 2020-06-18
Implement some light refactors to a few components

jameswburke:
* Splits out a new `irving/post-published-timestamp` from `irving/post-byline` and removes the timestamp from byline, simplifying its purpose.
* Refactors config naming for Social Sharing to be more in line with other components.

Screen Shot 2020-06-18 at 4 25 20 PM

### DIFF
--- a/inc/components/post-timestamp/component.php
+++ b/inc/components/post-timestamp/component.php
@@ -1,8 +1,8 @@
 <?php
 /**
- * Post published timestamp.
+ * Post timestamp.
  *
- * Get the post published timestamp.
+ * Get the post's timestamp(s).
  *
  * @package WP_Irving
  */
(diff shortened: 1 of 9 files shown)


inc/components/post-timestamp/component.php
```diff
@@ -0,0 +1,49 @@
...
+
+			return $component->set_config(
+				'content',
+				sprintf(
+					/**
+					 * Translators:
+					 * %1$s - Published timestamp.
+					 * %2$s - Modified timestamp.
+					 */
+					$component->get_config( 'content_format' ),
```
amyevans, line 42: Interesting, I like this way of doing it, def provides flexibility

jameswburke: I stumbled on the pattern building out the search. Gives the end user all the tools they need to build complex output without touching anything besides JSON. Pretty cool.

PR #177, opened 2020-06-19
IRV-385: Refactor nodes in admin bar

jameswburke:
* Moves Irving admin bar nodes from `class-admin.php` to `admin-bar.php`.
* Introduces nodes for more API helpers, settings, and docs.

### DIFF
--- a/inc/admin-bar.php
+++ b/inc/admin-bar.php
@@ -241,8 +241,8 @@ function get_admin_api_items(): array {
 	 */
 	$term = get_term_by(
 		'term_taxonomy_id',
-		absint( $_GET['tag_ID'] ?? 0 ), // phpcs:disable WordPress.Security.NonceVerification.Recommended
-		sanitize_text_field( wp_unslash( $_GET['taxonomy'] ?? '' ) ) // phpcs:disable WordPress.Security.NonceVerification.Recommended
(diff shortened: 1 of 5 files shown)


inc/admin-bar.php
```diff
@@ -54,7 +58,10 @@ function setup_admin_bar(
 			'irving/admin-bar',
 			[
 				'config' => [
-					'iframe_src' => site_url( $path ),
+					'iframe_src' => add_query_arg(
+						$endpoint->custom_params,
+						site_url( $path )
+					),
```
jameswburke, line 64: Needed to include the query params so that the admin bar accurately pulls in that context (ex. `?s=test`).

inc/admin-bar.php
```diff
@@ -127,3 +134,217 @@ function wp_head() {
...
+	if ( ! is_admin() ) {
+		return [];
+	}
+
+	$admin_api_items = [];
+
+	/**
+	 * Add a node for the post we're currently editing.
+	 */
+	$post_id = absint( $_GET['post'] ?? 0 ); // phpcs:ignore
```
pattiereaves, line 226: use the same phpcs disable syntax for consistency?

jameswburke: I want to use `ignore` actually. Updating now. Thx.

PR #180, opened 2020-06-24
Refactor context and registry into component instantiation

joemcgill:
This reorganizes the Component class so that on instantiation, a Component class automatically
accepts configuration as a param and returns a hydrated instance of itself and its children.

To achieve this goal, dependencies to the context store and component registry have been moved into
the Component class, creating an intentional hard coupling between these parts of the architecture
and all classes have been moved into the Component namespace to better signal that coupling.

To let registered components to dynamically alter the configuration or children values during the
hydration process, the callback process has been split into two distinct callbacks:

* config_callback – A callback that accepts in the current config and returns a hydrated config array.
* children_callback – A callback that accepts the current array of children and config values and returns a hydrated array of children.

This PR also includes a significant reorganization of namespaces and the file structure to reinforce the goals of separating templating and hydration concerns.

* All component related functionality now exists in the `inc/components/` directory and are moved into the `WP_Irving\Components` namespace.
* All template related functionality now exists in the `inc/templates/` directory and are moved into `WP_Irving\Templates` namespace.


inc/components/namespace.php
```diff
@@ -36,11 +68,11 @@ function auto_register_components() {
...
 		[
-			'wp_irving' => WP_IRVING_PATH . '/inc/components/', // Load components from WP Irving.
+			'wp_irving' => WP_IRVING_PATH . '/inc/components/components', // Load components from WP Irving.
 			'theme'     => get_stylesheet_directory() . '/components/', // Load components from the activated theme.
 		]
 	);
 
 	get_registry()->load_components( $directories );
 }
-add_filter( 'init', __NAMESPACE__ . '\auto_register_components' );
```
joemcgill, line 46: Moved to `bootstrap()` function.

inc/components/namespace.php
```diff
@@ -1,16 +1,22 @@
 <?php
```
joemcgill, line 1: This introduces a new file naming pattern, where functions of a namespace reside in a `namespace.php` file in the directory for that namespace. Additionally, all filters and actions get hooked from inside a `bootstrap()` function that is called from the main plugin file after the namespace files are loaded.

inc/components/namespace.php
```diff
@@ -22,6 +28,32 @@ function get_registry(): Registry {
...
+/**
+ * Returns the context object.
+ *
+ * Sets the default 'irving/post' context when first called.
+ *
+ * @return Context_Store The context store object.
+ */
+function get_context_store() {
+	global $wp_query;
+	global $wp_irving_context;
```
joemcgill, line 40: This needed to be turned into a global so that the defaults can be manipulated during Unit Testing. Unsure if there are other use cases where we would need to reset the defaults during a single application run, but that's why I switched it from a static value.

inc/templates/namespace.php
```diff
@@ -130,56 +130,62 @@ function get_template_path( WP_Query $query ): string {
...
-			sprintf(
-				// Translators: %1$s: Error message, %2$s Template path.
-				esc_html__( 'Error: %1$s found in %2$s.', 'wp-irving' ),
-				esc_html( json_last_error_msg() ),
-				esc_html( $template )
-			)
-		);
-	}
-
-	if ( has_blocks( $contents ) ) {
```
joemcgill, line 169: I want to revisit block parsing later when we look at Gutenberg integrations. Prioritizing JSON templates during this run.

tests/components/test-class-component.php
```diff
@@ -0,0 +1,624 @@
...
+			],
+			$component->get_config()
+		);
+	}
+
+	/**
+	 * Test children set from config.
+	 */
+	public function test_children_from_config() {
+		// Test both array and string notation.
```
joemcgill, line 436: Unsure if we really want to support both notation types or if we want to enforce the array syntax 🤷‍♂️

tests/templates/test-templates.php
```diff
@@ -75,28 +76,141 @@ public function test_template_paths( $paths ) {
...
+						[
+							'children' => [
+								'example/component3',
+							],
+						]
+					),
+					new Component( 'example/component' ),
+				],
+				'Nested template parts are not hydrating correctly.',
+			],
```
joemcgill, line 190: Are there other specific test cases that we want to add here?

tests/components/test-components/use-context.json
```diff
@@ -0,0 +1,17 @@
...
+    "foo": {
+      "default": "default",
+      "type": "string"
+    },
+    "baz": {
+      "default": "default",
+      "type": "string"
+    }
+  },
+	"use_context": {
```
jameswburke, line 13:
```suggestion
  "use_context": {
```

tests/templates/test-templates.php
jameswburke: Nested template parts?

tests/components/test-class-component.php
jameswburke: Can you elaborate on what dot notion in this situation is?

inc/components/class-component.php
```diff
@@ -828,20 +957,24 @@ public function jsonSerialize() {
...
-			if ( $config_schema['hidden'] ) {
-				$this->set_config( $key, null );
+		foreach ( $this->get_schema() as $key => $schema ) {
+			if ( $schema['hidden'] && isset( $config[ $key ] ) ) {
+				unset( $config[ $key ] );
 			}
 		}
 
+		// Rename to an alias if one is set.
+		$name = empty( $this->get_alias() ) ? $this->get_name() : $this->get_alias();
```
jameswburke, line 973: For the sake of debugging, it's a lot easier if we don't modify `name`, but instead have the `_alias` key. For example, this way we're not looking at `irving/text` when we should be debugging in `irving/post-title`.

I'm strongly in favor of leaving alias as its own key.

inc/templates/namespace.php
```diff
@@ -364,263 +368,58 @@ function convert_blocks_to_components( array $blocks ): array {
...
+			// If the component is converted to template data,
+			// add hydrated components to the array and move on.
+			if ( ! empty( $template_data ) ) {
+				array_push( $hydrated, ...$template_data );
+			}
 
-			// A little cleanup.
-			unset( $template_data );
-			break;
+			continue;
```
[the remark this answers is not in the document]
joemcgill: Yeah, there was another test that broke for the same reason.

inc/templates/namespace.php
```diff
@@ -364,263 +368,58 @@ function convert_blocks_to_components( array $blocks ): array {
...
 
-		// Set up config from context.
-		$component->use_context( get_template_context() );
-
-		// Run hydration callback.
-		$component->do_callback();
-
-		// Update context values.
-		$component->provide_context( get_template_context() );
+		$component = new Component( $component['name'], $component );
```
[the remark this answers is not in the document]
joemcgill: I'm unsure when this would even happen unless someone makes a mistake while writing a template file. I think I would rather throw an explicit warning/error message rather than hide/obscure the error so that it can more easily be identified and fixed.

inc/components/class-component.php
```diff
@@ -334,10 +406,80 @@ public function get_children(): array {
...
+
+			$name = $child[0];
+			$args = isset( $child[1] ) ? $child[1] : [];
+
+			// Replace the array with an initialized component instance.
+			$child = new Component( $name, $args );
+		}
+
+		if ( is_callable( $this->children_callback ) ) {
+			$new_children = call_user_func_array( $this->children_callback, [ $children, $this->config ] );
```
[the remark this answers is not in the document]
joemcgill: Yeah, this is the most difficult part of the whole process. We have to control the order of operations between setting config values and populating children in order to have a consistent flow of configuration data from ancestors to descendants during hydration. We could treat the current theme as a part of configuration, or even pass the whole component into the callback function, but I've avoided doing so to protect against someone modifying the object in unexpected ways during the callback process.

inc/components/class-component.php
joemcgill: I waffled on this a bit. My thinking is that it's more consistent for the front-end Irving app if all it really needs to worry about is mapping the `name` value of each component, rather than needing to worry about an alias property. For debugging, I think we could change the original name to a property value that is still passed to the front end though, like `_aliased_from` or something?

tests/templates/test-templates.php
joemcgill: Are you asking for an explanation of what a nested template part is? I was thinking of a case where you might have a template partial that references another template partial. We should support this use case and wanted to write a test that explicitly confirms that it's working.

tests/components/test-class-component.php
joemcgill: I was thinking it might be nice to support passing a list of component names as children, rather than making each be an array.

inc/components/class-component.php
jameswburke: That works for me.

PR #181, opened 2020-06-24
Hedron Grid support

jameswburke:


joemcgill:
@jameswburke is this still something we're wanting to add, or should we close this?

jameswburke:
We can kill this for now, but I'd like to talk about grids generally as part of a retro.

PR #185, opened 2020-07-02
Implement the video component

jameswburke:
* Implements a video component using oembed.
* Cleans up `irving/post-featured-media` and adds todos for a better UI later.

### DIFF
--- a/inc/components/post-featured-media/component.json
+++ b/inc/components/post-featured-media/component.json
@@ -8,11 +8,6 @@
 			"default": "",
 			"type": "string"
 		},
-    "caption": {
-      "description": "Optional. Caption for the image.",
-      "default": "",
-      "type": "string"
-    },
 		"object_fit": {
 			"default": "cover",
 			"description": "Object fit property.",
--- a/inc/components/post-featured-media/component.json
+++ b/inc/components/post-featured-media/component.json
@@ -8,6 +8,11 @@
 			"default": "",
 			"type": "string"
 		},
+    "caption": {
+      "description": "Optional. Caption for the image.",
+      "default": "",
+      "type": "string"
+    },
 		"object_fit": {
 			"default": "cover",
 			"description": "Object fit property.",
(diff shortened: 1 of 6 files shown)


inc/components/post-featured-media/component.json
```diff
@@ -8,6 +8,11 @@
 			"default": "",
 			"type": "string"
 		},
+    "caption": {
+      "description": "Optional. Caption for the image.",
```
amyevans, line 12: "Caption for the media", perhaps, to keep it more generic?

inc/components/video/component.php
```diff
@@ -0,0 +1,55 @@
...
+\WP_Irving\get_registry()->register_component_from_config(
+	__DIR__ . '/component',
+	[
+		'callback' => function( Component $component ): Component {
+
+			/**
+			 * If an aspect ratio value exists, check if it's an alias from the
+			 * mapping (or not), and use as the wrapper's padding-bottom
+			 * property. Values not aliased in the component map should be
+			 * complete values with a percetnage ex. "20%".
```
amyevans, line 30:
```suggestion
			 * complete values with a percentage ex. "20%".
```

inc/components/video/component.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "Render an HTML video embed responsively.",
+  "config": {
+    "aspect_ratio": {
+      "default": "16:9",
+      "description": "Required. Video ratio as a CSS property, ex. \"padding-bottom: `{aspect_ratio}%\". Alias options include `square`, `1;1`, `4:3`, `16:9`.",
+      "type": "string"
+    },
+    "video_url": {
+      "default": "",
+      "description": "Required. URL of the video. Currently supports YouTube and Vimeo.",
```
amyevans, line 13: Wouldn't basically anything listed in the providers list [here](https://developer.wordpress.org/reference/classes/wp_oembed/) work? (In which case, I think we should drop calling out certain providers by name here).

jameswburke: From what I can tell, the [React oEmbed Container package](https://github.com/humanmade/react-oembed-container) only supports some stuff, not everything that WP does. To avoid confusion/being misleading, I called out YouTube and Vimeo specifically..

PR #186, opened 2020-07-06
Begin updating core component callbacks after refactor.

joemcgill:
~**BLOCKED BY #180:** Change base branch of this PR after the former is merged.~

This is a followup to #180, which begins updating the core components to make use of the new `config_callback` and `children_callback` patterns so we can establish the pattern for the rest of the components to be fixed up.

All components have been cleaned up to use the correct namespaces and have consistent config property values, and are registered using the new `WP_Irving\Components\register_component_from_config()` helper method.

Components updated with unit tests include:

- [x] irving/archive-title
- [x] irving/post-byline
- [x] irving/post-content
- [x] irving/post-excerpt
- [x] irving/post-featured-image
- [x] irving/post-featured-media

Once we've established this process, we'll hit the rest of the components one-by-one, so this process isn't blocking other changes as we find and fix bugs and so future PRs are smaller and easier to review.


tests/components/test-components.php
```diff
@@ -17,6 +18,144 @@
...
+	public static $author_id;
+
+	/**
+	 * Test attachment
+	 *
+	 * @var int attachment ID.
+	 */
+	public static $attachment_id;
+
+
```
jameswburke, line 35:
```suggestion
```

inc/components/namespace.php
```diff
@@ -75,3 +75,83 @@ function auto_register_components() {
...
 	get_registry()->load_components( $directories );
 }
+
+/**
+ * Register a component.
+ *
+ * @param string $name Component name.
+ * @param array  $args Component args.
+ */
+function register_component( string $name, array $args = [] ) {
```
jameswburke, line 85:
```suggestion
function register_component( string $name, array $args = [] ): bool {
```

inc/components/namespace.php
```diff
@@ -75,3 +75,83 @@ function auto_register_components() {
 
 	get_registry()->load_components( $directories );
 }
+
+/**
+ * Register a component.
+ *
+ * @param string $name Component name.
+ * @param array  $args Component args.
```
jameswburke, line 83:
```suggestion
 * @param array  $args Component args.
 * @return bool Returns true if successful.
```

jameswburke: Feel free to update to whatever verbiage you prefer.

inc/components/namespace.php
```diff
@@ -75,3 +75,83 @@ function auto_register_components() {
...
+	}
+
+	return register_component(
+		$config['name'],
+		array_merge_recursive( $config, $args )
+	);
+}
+
+/**
+ * Determine an `alt` attribute value for this image.
```
jameswburke, line 140:
```suggestion
 * Determine an `alt` attribute value for a given image.
```

"this image" reads weird to me?

tests/components/test-components.php
```diff
@@ -47,4 +207,226 @@ public function test_template_default_context() {
...
+		return [
+			[
+				'irving/archive-title',
+				function () {
+					return [
+						'name'     => 'irving/archive-title',
+						'_alias'   => 'irving/text',
+						'config'   => (object) [
+							// The format of the archive title changed in WP version 5.5.
```
jameswburke, line 253: What a pita.

tests/components/test-components.php
```diff
@@ -47,4 +207,226 @@ public function test_template_default_context() {
...
+							'content'      => version_compare( get_bloginfo( 'version' ), '5.4.99', '>' ) ? 'Category: <span>Uncategorized</span>' : 'Category: Uncategorized',
```
jameswburke, line 254: Would `version_compare( get_bloginfo( 'version' ), '5.5.00', '>=' )` work/make more sense?

I understand the .99, and even though it's a sane, high value, is there some other version format that I'm unaware of that causes this to make more sense?

tests/components/test-components.php
joemcgill: I could make 100 gyros with all this pita 😆

tests/components/test-components.php
joemcgill: I tried it and it does not work. 5.4.99 works as expected so I decided to stop wasting time and take my PITA medicine.

joemcgill:
> ...is there a better way to test the components? I get the feeling that file is going to get very large soon.

Yeah, I think we might consider moving this to separate tests for each component type, rather than a single data provider. We could reuse the same assertion for each so we're not duplicating the same logic over again.

joemcgill:
@jameswburke I've split the component tests into separate methods and I think it's probably much cleaner.

tests/components/test-components.php
jameswburke: Fantastic.

PR #188, opened 2020-07-10
Refactor head component management

jameswburke:
IRV-512

* Refactors the `<Helmet>` logic into a `<Head>` component.
* Deprecates/removes Yoast integration code for WP Components.
* Updates unit tests.
* Introduces Util pattern and namespace.


inc/components/class-component.php
```diff
@@ -458,6 +457,17 @@ function ( $child ) {
...
+		/**
+		 * Filter the children values for all instances of this component.
+		 *
+		 * @todo Decide how this filter should be approached. This is a
+		 *       temporary solution that requires discussion.
+		 *
+		 * @param array $children Children for this component.
+		 * @param array $config   Config for this component.
+		 */
+		$children = apply_filters( 'wp_irving_component_children_' . $this->get_name(), $children, $this->config );
```
joemcgill, line 469: I think I'd rather have a general use filter here that passes the component name as a parameter rather than a dynamic one, sine this would result in integrators writing code that looks like this, which is fairly unorthodox: 

```
add_filters( 'wp_irving_component_chidren_irving/post-list', function () { ... } );
```

inc/components/components/head/component.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace WP_Irving\Components;
+
+/**
+ * Register the component and callback.
+ */
+register_component_from_config(
+	__DIR__ . '/component',
+	[
+		'children_callback' => function( $children, $config ): array {
```
joemcgill, line 18: You can add typecasting to the input params too.

inc/endpoints/class-components-registry-endpoint.php
```diff
@@ -24,7 +25,7 @@ public function register_rest_routes() {
 			[
 				'methods'  => \WP_REST_Server::READABLE,
 				'callback' => function() {
-					return get_registry()->get_registered_components();
+					return Components\get_registry()->get_registered_components();
```
joemcgill, line 28: 👍 Do we need a general helper function for this?

inc/templates/head-management.php
```diff
@@ -0,0 +1,87 @@
+<?php
+/**
+ * Automatically inject an `irving/head` component to manage the document
+ * <head>.
+ *
+ * @package WP_Irving
+ */
+
+namespace WP_Irving\Templates;
```
joemcgill, line 9: This should either be a sub-namespace, like `WP_Irving\Templates\Head_Management` or these functions should live in a general `inc/templates/namespace.php` file.

inc/templates/head-management.php
```diff
@@ -0,0 +1,87 @@
...
+ * @param array $children Children for the <head>.
+ * @return array
+ */
+function inject_favicon( array $children ): array {
+	return array_merge(
+		$children,
+		\WP_Irving\Utils\html_to_components( get_favicon_markup(), [ 'link', 'meta' ] )
+	);
+}
+add_action( 'wp_irving_defaults_head_component_children', __NAMESPACE__ . '\inject_favicon' );
```
joemcgill, line 87: Move this `add_action()` call into a `bootstrap()` function at the top of the file, and then call the bootstrap function after this file is loaded. See: https://github.com/alleyinteractive/wp-irving/blob/master/wp-irving.php#L77-L78

inc/utils/html-markup.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
```
joemcgill, line 2: Maybe this function should live in `inc/components/namespace.php` rather than creating a Utils namespace specifically for html-markup helpers?

tests/templates/test-head-management.php
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	/**
+	 * Setup.
+	 */
+	public function setUp() {
+		parent::setup();
+
+		// Disable date organized uploads.
+		add_filter( 'upload_dir', [ $this, 'upload_dir_no_subdir' ] );
```
joemcgill, line 73: Instead of duplicating this in several test classes, we should move this to a general helpers file that gets loaded during the bootstrapping process. Honestly, I'm unsure why the helper from the WP Unit tests wasn't working in the first place 🤷

tests/templates/test-head-management.php
```diff
@@ -0,0 +1,337 @@
...
+
+		// Run `setup_head()` on empty/blank data.
+		$component_endpoint_result = $this->get_head_setup_result();
+
+		// Nothing should have happened.
+		$this->assertEquals( [], $component_endpoint_result['defaults'], 'Defaults array was not empty.' );
+		$this->assertEquals( [], $component_endpoint_result['page'], 'Page array was not empty.' );
+
+		// Re-enable functionality.
+		add_filter( 'wp_irving_setup_head', '__return_true' );
```
joemcgill, line 178: I'm pretty sure this is unnecessary and the WP tests automatically reset hooks after each test, but you may want to confirm. 🔢

inc/components/class-component.php
jameswburke: Yeah, we need to discuss and figure out a plan of attack. Do you have any thoughts on it right now?

inc/endpoints/class-components-registry-endpoint.php
jameswburke: I don't think so, and if so, we can tackle it later - this is just fixing the namespace change after the refactor.

inc/components/class-component.php
joemcgill: I'd do exactly what you're doing, but name it generically, `wp_irving_component_children` with three params: `array $children, array $config, string, $name`. Placement of the filter is exactly where I was going to put it.

jameswburke:
@joemcgill, ready for another pass.

PR #189, opened 2020-07-10
Update post-list component

joemcgill:
This also fixes some bugs in the way children returned from a children callback were
being hydrated (i.e., they weren't). Updates all tests that were incorrect.

This may inadvertantly fix the text node support issues we were having with children as well.

### DIFF
--- a/tests/components/test-class-component.php
+++ b/tests/components/test-class-component.php
@@ -417,8 +417,9 @@ class Test_Class_Component extends WP_UnitTestCase {
 			'example/parent',
 			[
 				'children' => [
-					[ 'example/child-1' ],
-					'example/child-2',
+					'Text node.',
+					[ 'example/child' ],
+					[ 'name' => 'example/child' ],
 				],
 			]
 		);
@@ -426,8 +427,9 @@ class Test_Class_Component extends WP_UnitTestCase {
 		// Children should return as Component classes.
 		$this->assertEquals(
 			[
-				new Component( 'example/child-1' ),
-				new Component( 'example/child-2' ),
+				'Text node.',
+				new Component( 'example/child' ),
+				new Component( 'example/child' ),
 			],
 			$component->get_children()
 		);
--- a/tests/components/test-components.php
+++ b/tests/components/test-components.php
@@ -493,7 +493,7 @@ class Test_Components extends WP_UnitTestCase {
 									]
 								),
 							],
-						],
+						]
 					),
 					$this->get_expected_component( 'example/after' ),
 				],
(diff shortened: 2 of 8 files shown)


tests/components/test-class-component.php
```diff
@@ -417,17 +417,19 @@ public function test_children_from_config() {
 			'example/parent',
 			[
 				'children' => [
-					[ 'example/child-1' ],
-					'example/child-2',
+					'Text node.',
```
joemcgill, line 420: This test asserts that we are now supporting a couple different syntaxes when adding children.

1. Text nodes remain text.
2. An array with a single value can represent a component type.
3. An array with a `name` key will turn into a component of that type.

tests/components/test-components.php
```diff
@@ -173,12 +173,13 @@ public function setUp() {
...
-	 * @param string $message  Optional. Message for failure.
+	 * @param array     $expected     Callback to set the expected component after hydration.
+	 * @param Component $component Name of the component being tested.
+	 * @param string    $message      Optional. Message for failure.
+	 *
+	 * @see \PHPUnit\Framework\Assert::assertEquals
 	 */
-	public function assertComponentEquals( array $expected, string $name, string $message = '' ) {
-		$component = new Component( $name );
+	public function assertComponentEquals( array $expected, Component $component, string $message = '' ) {
```
joemcgill, line 182: It's easier to run assertions against an already instantiated component, rather than needing to pass in component arguments as well.

tests/components/test-components.php
```diff
@@ -431,6 +442,167 @@ public function test_component_post_featured_media() {
...
+				],
+				'children' => [ 'No results found.' ],
+			]
+		);
+
+		$this->assertComponentEquals( $expected, $component );
+	}
+
+	/**
+	 * Helper for creating expected output for components.
```
joemcgill, line 576: Got tired of writing the same defaults all the time. Maybe more confusing than it's worth?

tests/components/test-components.php
```diff
@@ -431,6 +442,167 @@ public function test_component_post_featured_media() {
...
+										'config'   => [
+											'postId' => $this->get_post_id(),
+										],
+										'children' => [
+											$this->get_expected_component( 'example/item' ),
+										],
+									]
+								),
+							],
+						],
```
jameswburke, line 496:
```suggestion
						]
```

jameswburke: Causing travis to fail.

joemcgill: Weird. Didn't see this on any of my local tests. Looks like we've got Travis configured to test against PHP versions 7.1 and 7.2. We should probably upgrade those to 7.3 and 7.4 and add a `PHPCompatibilityWP` to our sniffs.

PR #193, opened 2020-07-14
Allow short ternary operator

joemcgill:
This excludes the `WordPress.PHP.DisallowShortTernary` rule from the WP coding standards
and removes places where we've explicitly ignored this rule in our code.


joemcgill:
@amyevans Yeah, I think so, see: https://github.com/alleyinteractive/Alley-Coding-Standards/pull/6

PR #203, opened 2020-07-20
IRV-384: Add a "clear the cache" button to the Irving admin bar

tylerreckart:
This PR adds a "Clear The Cache" button to the WordPress admin bar for users that have the `manage_options` capability. On dev environments using this button will run the `fire_wipe_request()` method and on pantheon environments, the `pantheon_flush_site()` method will be targeted instead.

### DIFF
--- a/inc/endpoints/class-cache-endpoint.php
+++ b/inc/endpoints/class-cache-endpoint.php
@@ -13,15 +13,15 @@ namespace WP_Irving\REST_API;
 class Cache_Endpoint extends Endpoint {
 	/**
 	 * Attach to required hooks for endpoint.
-	*/
+	 */
 	public function __construct() {
 		parent::__construct();
 
 		// Register the route.
 		add_action( 'rest_api_init', [ $this, 'register_rest_routes' ] );
 		// Cache purge actions.
-		add_action( 'wp_irving_site_cache_purge', [ $this, 'purge_site_cache' ] );
-		add_action( 'wp_irving_page_cache_purge', [ $this, 'purge_page_cache' ] );
+		add_action( 'irving_site_cache_purge', [ $this, 'purge_site_cache' ] );
+		add_action( 'irving_page_cache_purge', [ $this, 'purge_page_cache' ] );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


inc/templates/admin-bar.php
```diff
@@ -348,3 +349,50 @@ function get_documentation_items(): array {
...
 	];
 }
+
+/**
+ * Get an array of Irving cache busting node args to be passed into
+ * admin_bar->add_node().
+ *
+ * @return array
+ */
+function get_cache_items(): array {
```
jameswburke, line 359: Is there a way to extend this to enable "Site Wide" on pantheon (where it's supported it), and also a "This page only" version for Patheon + VIP?

inc/templates/admin-bar.php
```diff
@@ -348,3 +349,50 @@ function get_documentation_items(): array {
...
+/**
+ * Register the click listener for the `irving-cache-purge` node. Register
+ * an action event, `wp_ajax_irving_cache_purge`, on click using WP ajax.
+ */
+function cache_purge_click_listener() {
+	?>
+		<script type="text/javascript">
+			jQuery('#wp-admin-bar-irving-cache-purge').on('click', function() {
+				const data = { 'action': 'irving_cache_purge' };
+				jQuery.post( '<?php echo esc_url( admin_url( 'admin-ajax.php' ) ); ?>', data );
```
jameswburke, line 378: Alley standards prefer REST API over `admin-ajax.php`, [adding a new endpoint](https://github.com/alleyinteractive/wp-irving/tree/master/inc/endpoints) would be preferred, and this function would call that instead.

inc/templates/admin-bar.php
```diff
@@ -348,3 +349,50 @@ function get_documentation_items(): array {
...
+		</script>
+	<?php
+}
+add_action( 'admin_footer', __NAMESPACE__ . '\cache_purge_click_listener', 95 );
+
+/**
+ * Purge the cache. Called when the `wp_ajax_irving_cache_purge` action
+ * is fired.
+ */
+function cache_purge_callback() {
```
jameswburke, line 389: Re: above, this logic should move into an endpoint, and it'd be great to split it up into "clear everything" & "clear this".

tylerreckart: Interesting thought. I'll dig into that!

inc/endpoints/class-cache-endpoint.php
```diff
@@ -0,0 +1,97 @@
...
+			\WP_Irving\Cache::instance()->fire_wipe_request();
+		}
+	}
+
+	/**
+	 * Purge the cache for a specific route/page.
+	 *
+	 * @param string $route The target route to purge from the cache.
+	 */
+	public function purge_page_cache( $route ) {
```
ostowe, line 86: Looks like maybe this might not yet account for purging page-specific endpoints and/or Redis cache, is that right? not necessarily something you need to include in this PR, but might make a note of it if that is the case.

tylerreckart: 🤔  Yeah that is something that I hadn't considered. I'll make a new ticket to add that to a future update.

PR #204, opened 2020-07-20
IRV-554: Update the menu component

joemcgill:
* Updates registration to use the 'config_callback' pattern
* Updates format of component.json for config items
* Adds unit test
* Updates CS rules/tools

### DIFF
--- a/inc/components/components/site-menu/component.json
+++ b/inc/components/components/site-menu/component.json
@@ -13,12 +13,14 @@
       "default": false
     },
     "location": {
-      "type": "string",
-      "required": true
+      "type": "string"
+    },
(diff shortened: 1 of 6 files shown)


inc/components/components/site-menu/component.json
```diff
@@ -13,12 +13,14 @@
       "default": false
     },
     "location": {
-      "type": "string",
-      "required": true
+      "type": "string"
```
jameswburke, line 16: Suggesting this since $config[''location'] is used below without any validation that it exists first, and overall the component is pretty useless without it.

```suggestion
      "default": "",
      "type": "string"
```

PR #208, opened 2020-07-21
Resolve template parts as component children

joemcgill:
This fixes a bug where template-parts that were referenced as children of a component were not being resolved and converted to components during hydration.


joemcgill:
Thanks @jameswburke. I've added some logic to `hydrate_template()` that should fix this issue, though I'm also not opposed to us documenting that template-parts should always return an array of component objects, to avoid this problem altogether.

PR #210, opened 2020-07-24
IRV-384: fix cache clearing behavior

tylerreckart:


inc/endpoints/class-cache-endpoint.php
```diff
@@ -51,8 +51,12 @@ public function get_route_response( $request ) {
 
 		// Ensure the action exists prior to firing.
 		if ( ! empty( $action ) ) {
+			$response = new \WP_REST_Response( true );
```
dlh01, line 54: Since, I assume, the server has a better handle on what happened with the cache than the client, perhaps the "success" string could be used as the response data directly here, and the JS then reads the string? Here might also be a more-natural place to pass the string to a translation function, which is currently missing.

inc/endpoints/class-cache-endpoint.php
```diff
@@ -62,9 +66,15 @@ public function get_route_response( $request ) {
 				// to firing the action.
 				if ( ! empty( $route ) ) {
 					do_action( 'wp_irving_page_cache_purge', $route );
+
+					return $response;
 				}
 			}
 		}
+
+		$response = new \WP_REST_Response( false );
```
dlh01, line 75: It might be worth returning a `WP_Error` object in this case with an error code and the error string from the JS, rather than a `WP_REST_Response`. The error will automatically be converted to a REST response, and WordPress will automatically set the HTTP response code is `500`, which might be more appropriate (a different HTTP response code can also be passed to the `WP_Error`). The downside is that `formatAlert()` will be a little more complicated because it needs to check for the error response.

inc/templates/admin-bar.php
```diff
@@ -392,16 +394,31 @@ function cache_purge_click_listener() {
 	$rest_endpoint = site_url( '/wp-json/irving/v1/purge-cache' );
 	?>
 		<script type="text/javascript">
+			function formatAlert(response) {
+				if (response === true) {
+					return 'Cache purged sucessfully';
+				}
+				return 'There was an error clearing the cache.\nPlease check the network tab in the developer console for more information.';
```
dlh01, line 401: One of these strings ends with punctuation, and one doesn't.

inc/endpoints/class-cache-endpoint.php
tylerreckart: @dlh01 Definitely helpful suggestions. Thanks!

PR #214, opened 2020-08-04
Pico integration

jameswburke:
See https://alleyinteractive.atlassian.net/browse/IRV-526

Adds a Pico integration to the integrations manager.

Used in parallel with https://github.com/alleyinteractive/irving/pull/251.

### DIFF
--- a/inc/integrations/class-pico.php
+++ b/inc/integrations/class-pico.php
@@ -7,7 +7,7 @@
 
 namespace WP_Irving\Integrations;
 
-use WP_Irving\Components\Component;
+use WP_Irving\Singleton;
 use Pico_Setup;
 use Pico_Widget;
 
@@ -15,6 +15,7 @@ use Pico_Widget;
  * Class to integrate Pico with Irving.
  */
 class Pico {
+	use Singleton;
 
 	/**
 	 * Setup the integration.
--- a/inc/integrations/class-pico.php
+++ b/inc/integrations/class-pico.php
@@ -5,60 +5,64 @@
  * @package WP_Irving
  */
 
-namespace WP_Irving;
+namespace WP_Irving\Integrations;
 
 use WP_Irving\Components\Component;
 use Pico_Setup;
 use Pico_Widget;
 
 /**
- * Pico.
+ * Class to integrate Pico with Irving.
  */
 class Pico {
 
 	/**
-	 * Constructor for class.
+	 * Class instance.
+	 *
+	 * @var null|self
 	 */
-	public function __construct() {
+	protected static $instance;
 
-		// Ensure Yoast exists and is enabled.
+	/**
+	 * Get class instance.
+	 *
+	 * @return self
+	 */
+	public static function instance() {
+		if ( ! isset( static::$instance ) ) {
+			static::$instance = new static();
+		}
+		return static::$instance;
+	}
+
+	/**
+	 * Setup the singleton. Validate JWT is installed, and setup hooks.
+	 */
+	public function setup() {
+
+		// Ensure Pico exists and is enabled.
 		if ( ! defined( 'PICO_VERSION' ) ) {
 			return;
 		}
 
 		if ( ! is_admin() ) {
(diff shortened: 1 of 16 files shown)


inc/integrations/class-pico.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * WP Irving integration for Pico.
+ *
+ * @package WP_Irving
+ */
+
+namespace WP_Irving\Integrations;
+
+use WP_Irving\Components\Component;
```
joemcgill, line 10: This appears unused, can be removed.

inc/integrations/class-pico.php
```diff
@@ -0,0 +1,61 @@
...
+use Pico_Setup;
+use Pico_Widget;
+
+/**
+ * Class to integrate Pico with Irving.
+ */
+class Pico {
+
+	/**
```
joemcgill, line 19: The rest of these integrations use a singleton pattern, though it looks like our `load_integrations()` function is instantiating a new class for each integration instead of getting a static instance. Can we fix this real quick?

inc/templates/namespace.php
```diff
@@ -530,12 +530,36 @@ function setup_head(
 /**
  * Manage any active integrations by inserting an `irving/integrations` component.
  *
- * @param array $data Data object to be hydrated by template.
- * @return array The update endpoint data.
+ * @param array     $data    Data object to be hydrated by templates.
+ * @param \WP_Query $query   The current WP_Query object.
```
joemcgill, line 534: We have a use statement for WP_Query, so the slash is unnecessary.

```suggestion
 * @param WP_Query $query   The current WP_Query object.
```

inc/templates/namespace.php
```diff
@@ -530,12 +530,36 @@ function setup_head(
...
+ * @param string    $context The context for this request.
+ * @return array The updated endpoint data.
  */
-function setup_integrations( array $data ): array {
+function setup_integrations( array $data, \WP_Query $query, string $context ): array {
```
joemcgill, line 538: Same as above.

```suggestion
function setup_integrations( array $data, WP_Query $query, string $context ): array {
```

inc/integrations/class-pico.php
jameswburke: Ah, that's why I dropped the singleton stuff from this class, I didn't realize that was the intention. I'll push an update.

jameswburke: Added a singleton and updated.

inc/integrations/namespace.php
```diff
@@ -56,6 +55,7 @@ function auto_register_plugin_integrations() {
 	$integrations = [
 		'jwt_auth' => __NAMESPACE__ . '\JWT_Auth',
 		'pico'     => __NAMESPACE__ . '\Pico',
+		'Yooo'     => __NAMESPACE__ . '\Whatever',
```
joemcgill, line 58: 😂 
```suggestion
```

inc/integrations/namespace.php
```diff
@@ -72,7 +72,14 @@ function auto_register_plugin_integrations() {
...
-		$integration->setup();
+
+		// Ensure the integration exists, and has an `instance` method using a
+		// singleton pattern.
+		if ( ! class_exists( $class ) || ! method_exists( $class, 'instance' ) ) {
+			continue;
+		}
+
+		// Create a singleton instance of this integration.
+		$integration = call_user_func( [ $class, 'instance' ] );
```
joemcgill, line 83: This could be simplified a bit.

```suggestion
		// Ensure the integration exists and has an instance method.
		if ( is_callable( [ $class, 'instance' ] ) ) {
			// Create a singleton instance of this integration.
			$integration = call_user_func( [ $class, 'instance' ] );
		}

```

inc/trait-singleton.php
```diff
@@ -0,0 +1,42 @@
...
+			static::$instance = new static();
+			static::$instance->setup();
+		}
+		return static::$instance;
+	}
+
+	/**
+	 * Setup the singleton.
+	 */
+	public function setup() {
```
joemcgill, line 39: ❤️

PR #215, opened 2020-08-04
IRV-592: Integrations Manager

tylerreckart:
**Important:** This PR should be reviewed in tandem with alleyinteractive/irving#258, as it includes core functionality on the React side that is required for this work to insert configured integrations into the DOM.

This PR is for the back-end work outlined in [IRV-592](https://alleyinteractive.atlassian.net/browse/IRV-592), which includes the creation and configuration of the "Irving Integrations" WP settings page, the creation of the `Google_Analytics` integration and the addition of the Tracking ID settings field to the integrations settings page, and the addition of the `irving/integrations` object in the Irving REST API to pass any configured fields to the integrations manager component on the front-end.

### DIFF
--- a/inc/integrations/class-google-analytics.php
+++ b/inc/integrations/class-google-analytics.php
@@ -74,8 +74,6 @@ class Google_Analytics {
 
 	/**
 	 * Render an input for the GA Tracking ID.
-	 *
-	 * @param array $args Arguments for input.
 	 */
 	public function render_tracking_id_input() {
 		// Check to see if there is an existing GA configuration in the option.
--- a/inc/integrations/class-google-analytics.php
+++ b/inc/integrations/class-google-analytics.php
@@ -68,10 +68,7 @@ class Google_Analytics {
 			esc_html__( 'Google Analytics Tracking ID', 'wp-irving' ),
 			[ $this, 'render_tracking_id_input' ],
 			'wp_irving_integrations',
-			'irving_integrations_settings',
-			[
-				'id' => 'tracking_id',
-			]
+			'irving_integrations_settings'
 		);
 	}
 
@@ -80,12 +77,12 @@ class Google_Analytics {
 	 *
 	 * @param array $args Arguments for input.
 	 */
-	public function render_tracking_id_input( array $args ) {
+	public function render_tracking_id_input() {
 		// Check to see if there is an existing GA configuration in the option.
-		$ga_key = $this->options['google_analytics'][ $args['id'] ] ?? '';
+		$ga_key = $this->options[ $this->option_key ]['tracking_id'] ?? '';
 
 		?>
-			<input type="text" name="irving_integrations[<?php echo esc_attr( 'ga_' . $args['id'] ); ?>]" value="<?php echo esc_attr( $ga_key ); ?>" />
+			<input type="text" name="irving_integrations[<?php echo esc_attr( 'ga_tracking_id' ); ?>]" value="<?php echo esc_attr( $ga_key ); ?>" />
 		<?php
 	}
 
--- a/inc/integrations/class-google-analytics.php
+++ b/inc/integrations/class-google-analytics.php
@@ -14,6 +14,8 @@ class Google_Analytics {
 
 	/**
 	 * Holds the option values to be set.
+	 *
+	 * @var array
 	 */
 	private $options;
 
@@ -27,7 +29,7 @@ class Google_Analytics {
 		// Register settings fields for integrations.
 		add_action( 'admin_init', [ $this, 'register_settings_fields' ], 10, 2 );
 		// Filter the updated option values prior to submission.
-		add_filter( 'pre_update_option_irving_integrations', [ $this, 'format_option_for_update' ], 10, 2);
+		add_filter( 'pre_update_option_irving_integrations', [ $this, 'format_option_for_update' ], 10, 2 );
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


inc/integrations/class-google-analytics.php
```diff
@@ -0,0 +1,94 @@
...
+		foreach ( $options as $option_key => $option_val ) {
+			// Build the config array for Google Analytics.
+			if ( strpos( $option_key, 'ga_' ) !== false ) {
+				$arr['google_analytics'][ str_replace( 'ga_', '', $option_key ) ] = $option_val;
+			}
+		}
+
+		return $arr;
+	}
+
```
jameswburke, line 86:
```suggestion
```

inc/integrations/class-google-analytics.php
```diff
@@ -0,0 +1,94 @@
...
+		foreach ( $options as $option_key => $option_val ) {
+			// Build the config array for Google Analytics.
+			if ( strpos( $option_key, 'ga_' ) !== false ) {
+				$arr['google_analytics'][ str_replace( 'ga_', '', $option_key ) ] = $option_val;
+			}
+		}
+
+		return $arr;
+	}
```
jameswburke, line 85: This part is a little fuzzy to me, can you add some context about why it's doing what it's doing? It seems like every integration would need to implement something similar?

inc/integrations/class-google-analytics.php
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * Group the updated options based on the integration prefix.
+	 *
+	 * @param array $options The updated options.
+	 * @return array $arr The formatted options.
+	 */
+	public function format_option_for_update( $options ): array {
+		// Construct an empty array.
+		$arr = [];
```
jameswburke, line 75: Can we name it something more explicit? I know it's an array, but what does it do?

inc/integrations/class-integrations-manager.php
```diff
@@ -0,0 +1,111 @@
...
+				<h1 class="wp-heading-inline">
+					<?php esc_html_e( 'WP-Irving - Integrations Manager', 'wp-irving' ); ?>
+				</h1>
+
+				<hr class="wp-header-end">
+
+				<form method="post" action="options.php">
+					<?php settings_fields( 'wp_irving_integrations' ); ?>
+					<?php do_settings_sections( 'wp_irving_integrations' ); ?>
+					<?php submit_button( __( 'Save Settings', 'wp-irving' ) ); ?>
```
joemcgill, line 99: 👍 Text value is escaped in `get_submit_button()` so no need to add additional escaping here, translation only is fine.

inc/integrations/class-integrations-manager.php
```diff
@@ -0,0 +1,111 @@
...
+				</form>
+			</div>
+		<?php
+	}
+}
+
+add_action(
```
joemcgill, line 106: Similarly, to above, let's move side effects out of the class file and into whatever file is including the integrations manager.

inc/integrations/class-integrations-manager.php
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	/**
+	 * Class constructor.
+	 */
+	public function setup() {
+		// Register admin page.
+		add_action( 'admin_menu', [ $this, 'register_admin' ] );
+		// Register settings fields for integrations.
+		add_action( 'admin_init', [ $this, 'register_settings_fields' ], 10, 2 );
```
joemcgill, line 42: `register_settings_fields()` doesn't accept any params, so you can omit the last two arguments here.

```suggestion
		add_action( 'admin_init', [ $this, 'register_settings_fields' ] );
```

inc/templates/namespace.php
```diff
@@ -524,6 +527,44 @@ function setup_head(
...
+ *
+ * @param array $data Data object to be hydrated by template.
+ * @return array The update endpoint data.
+ */
+function setup_integrations( $data ) {
+	// Get any set integrations options.
+	$options = get_option( 'irving_integrations' );
+
+	// If options are present, append the integrations component into the endpoint response.
+	if ( ! empty( $options ) ) {
```
joemcgill, line 541: Totally not a blocker, but this would read more clearly and avoid a nesting level by returning early here if `$options` is empty, i.e.

```php
// Bail early if no options are present.
if ( empty( $options ) ) {
	return $data;
}

// The rest of the logic here...
```

tylerreckart:
@joemcgill @jameswburke Thanks for the great feedback! Your suggestions have been implemented. Joe, let me know if the instance instantiation pattern I choose was along the lines of what you were thinking. All integration setup is now executed by the manager outside of the instance creation as you suggested. I also refactored the existing integration plugins to use the instance pattern so that the hooks were not being called in the `__construct()` function.

inc/integrations/class-integrations-manager.php
```diff
@@ -27,19 +30,82 @@ class Integrations_Manager {
...
 	 */
-	public function setup() {
+	public function setup_integrations() {
 		// Register admin page.
 		add_action( 'admin_menu', [ $this, 'register_admin' ] );
 		// Register settings fields for integrations.
-		add_action( 'admin_init', [ $this, 'register_settings_fields' ], 10, 2 );
+		add_action( 'admin_init', [ $this, 'register_settings_fields' ] );
+
+		// Archiveless.
```
joemcgill, line 46: This totally works for now. Could be cool to have an array of integration classes that are supported out of the box and then have one place that loops through all enabled integrations and loads them, something like this (pseudocode):

```php
$integrations = [
	'archiveless' => __NAMESPACE__ . '\\Archiveless',
	'google_analytics' => __NAMESPACE__ . '\\ Google_Analytics',
	//etc.
];

// Allow new integrations to be added via a filter.
$integrations = apply_filters( 'wp_irving_active_integrations', $integrations );

foreach ( $integrations as $type => $class ) {
	$integration = new $class();
	$integration->setup();
}
```

That would DRY out this method a bit and allow for projects to add custom integrations without Irving Core providing them. We could even write an interface for Integrations to help teams conform to the requirements of the manager.

inc/integrations/class-vip-go.php
```diff
@@ -65,6 +84,6 @@ function( $url ) {
 add_action(
```
joemcgill, line 84: Can we move this action call as well?

wp-irving.php
```diff
@@ -43,6 +42,24 @@
 require_once WP_IRVING_PATH . '/inc/integrations/class-pantheon.php';
 require_once WP_IRVING_PATH . '/inc/integrations/class-wpcom-legacy-redirector.php';
 require_once WP_IRVING_PATH . '/inc/integrations/class-yoast.php';
+// Integrations Manager.
+require_once WP_IRVING_PATH . '/inc/integrations/class-integrations-manager.php';
+// Instantiate the integrations manager.
+new \WP_Irving\Integrations\Integrations_Manager();
```
joemcgill, line 48: Let's create a new file at `/inc/integrations/namespace.php` which includes a bootstrap function that runs all of this code, rather than doing it in the main plugin file. See the pattern set up below for `Components` and `Templates`.

PR #217, opened 2020-08-06
Ensure all endpoints have a permissions callback

jameswburke:
WordPress 5.5 will start triggering `_doing_it_wrong` errors for any REST endpoints that don't set a permission_callback. See: https://make.wordpress.org/core/2020/07/22/rest-api-changes-in-wordpress-5-5/

To avoid these errors, we need to ensure all registered routes in the plugin include a permissions callback when registering the route.

### DIFF
--- a/inc/endpoints/class-cache-endpoint.php
+++ b/inc/endpoints/class-cache-endpoint.php
@@ -50,8 +50,8 @@ class Cache_Endpoint extends Endpoint {
 		/**
 		 * Filter the permissions check.
 		 *
-		 * @param bool|\WP_Error   $retval  Returned value.
-		 * @param WP_REST_Request $request The request sent to the API.
+		 * @param bool|callable   $permission_callback Filtered `permission_callback` value.
+		 * @param WP_REST_Request $request             Endpoint request.
 		 */
 		return apply_filters( 'wp_irving_cache_endpoint_permissions_check', current_user_can( 'manage_options' ), $request );
 	}
--- a/inc/endpoints/class-cache-endpoint.php
+++ b/inc/endpoints/class-cache-endpoint.php
@@ -32,8 +32,9 @@ class Cache_Endpoint extends Endpoint {
 			self::get_namespace(),
 			'/purge-cache',
 			[
-				'methods'  => \WP_REST_Server::CREATABLE,
-				'callback' => [ $this, 'get_route_response' ],
+				'callback'            => [ $this, 'get_route_response' ],
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'permission_callback' => '__return_true',
 			]
 		);
 	}
--- a/inc/endpoints/class-cache-endpoint.php
+++ b/inc/endpoints/class-cache-endpoint.php
@@ -34,9 +34,26 @@ class Cache_Endpoint extends Endpoint {
 			[
 				'callback'            => [ $this, 'get_route_response' ],
 				'methods'             => \WP_REST_Server::CREATABLE,
-				'permission_callback' => '__return_true',
+				'permission_callback' => [ $this, 'permissions_check' ],
 			]
 		);
+	}
+
+	/**
+	 * Permissions check.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return bool|\WP_Error
+	 */
+	public function permissions_check( $request ) {
+
+		/**
+		 * Filter the permissions check.
+		 *
+		 * @param bool|\WP_Error   $retval  Returned value.
+		 * @param WP_REST_Request $request The request sent to the API.
+		 */
+		return apply_filters( 'wp_irving_cache_endpoint_permissions_check', current_user_can( 'manage_options' ), $request );
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


inc/endpoints/class-cache-endpoint.php
```diff
@@ -32,8 +32,9 @@ public function register_rest_routes() {
 			self::get_namespace(),
 			'/purge-cache',
 			[
-				'methods'  => \WP_REST_Server::CREATABLE,
-				'callback' => [ $this, 'get_route_response' ],
+				'callback'            => [ $this, 'get_route_response' ],
+				'methods'             => \WP_REST_Server::CREATABLE,
+				'permission_callback' => '__return_true',
```
joemcgill, line 37: We should check for a capability in order to allow someone to clear the cache here. Maybe `current_user_can( 'manage_options' )` unless we want to give this capability to non-admins.

Otherwise, anyone in the world could send a POST request to an Irving endpoint and cause a cache stampede or construct a DoS attack.

inc/endpoints/class-data-endpoint.php
```diff
@@ -46,8 +46,9 @@ public function register_rest_routes() {
 				self::get_namespace(),
 				'/data/' . $endpoint['slug'],
 				[
-					'methods'  => \WP_REST_Server::READABLE,
-					'callback' => $endpoint['callback'],
+					'callback'            => $endpoint['callback'],
+					'methods'             => \WP_REST_Server::READABLE,
+					'permission_callback' => '__return_true',
```
joemcgill, line 51: Not sure how data endpoints are being used by integrators, but I wonder if we should use a similar pattern to how callbacks are registered and set the `permission_callback` to `$endpoint['permission_callback']` so integrations can supply their own permissions callback.

inc/endpoints/class-form-endpoint.php
```diff
@@ -46,8 +46,9 @@ public function register_rest_routes() {
 				self::get_namespace(),
 				'/form/' . $endpoint['slug'],
 				[
-					'methods'  => \WP_REST_Server::CREATABLE,
-					'callback' => $endpoint['callback'],
+					'callback'            => $endpoint['callback'],
+					'methods'             => \WP_REST_Server::CREATABLE,
+					'permission_callback' => '__return_true',
```
joemcgill, line 51: Similar to the caching endpoint, I don't think POST endpoints should ever be public, but I'm unsure what the correct capabilities should be for this endpoint.

jameswburke: This endpoint is used for public POSTs to forms, ex. contact form. Being public is expected.

joemcgill: Ah, thanks. My misunderstanding.

jameswburke:
@joemcgill updated to ensure that `permission_callback`s are filtered as we discussed, also added some unit tests for the permissions. I have a feeling they can be approached better, so let me know if you have any feedback there in particular.

inc/endpoints/class-data-endpoint.php
```diff
@@ -41,15 +42,29 @@ public function register_rest_routes() {
...
+					'callback'            => '__return_true',
+					'methods'             => \WP_REST_Server::READABLE,
+					'permission_callback' => '__return_true',
+					'slug'                => '',
 				]
 			);
+
+			// Ensure we have a slug.
+			if ( empty( $args['slug'] ) || ! is_string( $args['slug'] ) ) {
+				break;
```
joemcgill, line 60: This should be `continue;` otherwise, other endpoints in the array won't get registered.

tests/endpoints/test-class-cache-endpoint.php
```diff
@@ -0,0 +1,117 @@
...
+		wp_set_current_user( $subscriber->ID );
+
+		$this->assertEquals(
+			false,
+			$endpoint->permissions_check( $this->get_cache_endpoint_response() ),
+			'Permission check for subscriber should have failed.'
+		);
+
+		// Create and get a user with the administrator role.
+		$administrator = $this->factory->user->create_and_get();
```
joemcgill, line 81: You can set the role when you create a user with this factory, like:

```
$admin = $this->factory->user->create_and_get(
	[
		'role'       => 'administrator',
	]
);
```

`factory->{object}->create_and_get()` generally accepts all parameters used by `wp_insert_{object}()`, in this case: [wp_insert_user()](https://developer.wordpress.org/reference/functions/wp_insert_user/)

tests/endpoints/test-class-data-endpoint.php
```diff
@@ -0,0 +1,91 @@
...
+ * Test the data endpoint functionality.
+ *
+ * @group endpoints
+ */
+class Test_Class_Data_Endpoint extends WP_UnitTestCase {
+
+	/**
+	 * Class setup.
+	 */
+	public static function wpSetUpBeforeClass() {
```
joemcgill, line 26: I think this needs to use the `setUp()` method or else additional tests will not have this filter set due to hooks being reset after each test is run. It's fine for now since there is only one test method.

tests/endpoints/test-class-cache-endpoint.php
jameswburke: Thanks, is there a solid set of docs for unit tests? I can't find anything reliable.

tests/endpoints/test-class-data-endpoint.php
jameswburke: Good call. Thanks.

jameswburke:
@joemcgill ready for you again.

PR #218, opened 2020-08-11
IRV-615: Set yoast schema through integrations manager

tylerreckart:


inc/integrations/class-yoast.php
```diff
@@ -46,9 +47,45 @@ public function setup() {
...
+
+		foreach ( $components as $component ) {
+			$component_config   = $component->get_config();
+			$component_children = $component->get_children();
+
+			if (
+				isset( $component_config['type'] )
+				&& 'application/ld+json' === $component_config['type']
+			) {
+				$schema = $component->get_children()[0];
```
joemcgill, line 74: Assuming there is only going to be one of these, we can add a `break;` after this is found and stop looping through the rest of the components.

I also think if would probably be faster to do a regex lookup for the markup you're looking for from the head markup and parse out the content rather than converting everything to components first.

Something like this:

```php
preg_match( 
	'/<script type='application\/ld\+json'[^>]+>(.+)<\/script>/',
	$this->get_yoasts_head_markup(),
	$matches
);

$content = $matches[1] ?? "";

if ( ! empty( $content ) ) {
	$config[] = [ 'yoast_schema' => [ 'content' => $content ] ];
}
```

inc/integrations/class-yoast.php
```diff
@@ -46,9 +47,45 @@ public function setup() {
 		if ( ! is_admin() ) {
 			// Parse Yoast's head markup and inject it into the Head component.
 			add_filter( 'wp_irving_component_children', [ $this, 'inject_yoast_tags_into_head_children' ], 10, 3 );
+			add_filter( 'wp_irving_integrations_config', [ $this, 'inject_yoast_schema_into_integrations_config' ], 10, 1 );
```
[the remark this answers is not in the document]
tylerreckart: 👍

inc/integrations/class-yoast.php
tylerreckart: Great suggestion @joemcgill!

PR #219, opened 2020-08-13
Add config filter and fix bug in post-list component

jameswburke:
As titled.

### DIFF
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -549,7 +549,7 @@ class Component implements JsonSerializable {
 		 * @param array  $config Config for this component.
 		 * @param string $name   Name of this component.
 		 */
-		$new_config = apply_filters( 'wp_irving_component_config', $this->get_config(), $this->get_name() );
+		$new_config = apply_filters( 'wp_irving_component_config_pre_hydration', $this->get_config(), $this->get_name() );
 
 		if ( is_callable( $this->config_callback ) ) {
 			$new_config = call_user_func_array( $this->config_callback, [ $new_config ] );
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -544,7 +544,18 @@ class Component implements JsonSerializable {
 	 */
 	private function hydrate_config(): self {
 		if ( is_callable( $this->config_callback ) ) {
-			$new_config = call_user_func_array( $this->config_callback, [ $this->get_config() ] );
+
+			$current_config = $this->get_config();
+
+			/**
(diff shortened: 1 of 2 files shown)


inc/components/class-component.php
```diff
@@ -543,13 +543,21 @@ public function append_child( $child ): self {
 	 * @return self
 	 */
 	private function hydrate_config(): self {
+		/**
+		 * Filter the config values before the config callback fires.
+		 *
+		 * @param array  $config Config for this component.
+		 * @param string $name   Name of this component.
+		 */
+		$new_config = apply_filters( 'wp_irving_component_config', $this->get_config(), $this->get_name() );
```
joemcgill, line 552: Let's rename this something like `wp_irving_component_config_pre_hydration` or `wp_irving_component_config_data` so we can save `wp_irving_component_config` as a hook that fires after hydration—consistent with the `wp_irving_component_children` hook.

inc/components/class-component.php
```diff
@@ -543,13 +543,21 @@ public function append_child( $child ): self {
...
 		if ( is_callable( $this->config_callback ) ) {
-			$new_config = call_user_func_array( $this->config_callback, [ $this->get_config() ] );
+			$new_config = call_user_func_array( $this->config_callback, [ $new_config ] );
+		}
 
-			// @todo Add error handling.
-			if ( is_array( $new_config ) ) {
-				$this->set_config( $new_config );
-			}
+		// @todo Add error handling.
```
joemcgill, line 558: 👏

inc/components/components/post-list/component.php
```diff
@@ -57,7 +57,9 @@
 			}
 
 			$post_ids = wp_list_pluck( $query->posts, 'ID' );
-			$post_ids = post_list_get_and_add_used_post_ids( $post_ids );
+
+			// Add the current $post_ids to the list of used ids.
+			post_list_get_and_add_used_post_ids( $post_ids );
```
joemcgill, line 62: Ah! I hadn't noticed how this feature worked. This could cause lots of confusion since it persists throughout the entire page. Separating adding and getting used post IDs might be useful, along with making this more general use outside just post-lists so an individual post or component could add their ID to the queue. No need to do anything here, I'll open a new ticket.

jameswburke: You may be interested in this, https://github.com/alleyinteractive/alleypack/tree/master/modules/unique-wp-query/1.0

jameswburke: Also, agreed, it could be done better.

PR #221, opened 2020-08-14
post meta component

xmattus:


inc/components/components/post-meta/component.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "irving/post-meta",
+  "_alias": "irving/fragment",
+  "description": "Provide context for a specific post meta value.",
+  "config": {
+    "meta_key": {
+      "hidden": false,
```
joemcgill, line 7: `hidden` defaults to `false`, so these can be omitted.

inc/components/components/post-meta/component.json
```diff
@@ -0,0 +1,21 @@
...
+      "type": "string"
+    },
+    "meta_value": {
+      "hidden": false,
+      "type": "string"
```
joemcgill, line 12: Even though numbers get converted to strings, meta values could also be arrays, so we probably don't want to enforce a type here.

xmattus: I wasn't sure if the type was optional. If I just leave it out, will the plugin just skip type verification/coercion?

PR #223, opened 2020-08-14
Allow use_context in templates

xmattus:


xmattus:
@joemcgill Cool, looks great. Thanks!

PR #225, opened 2020-08-16
Add interstitial support in the post list component

jameswburke:
## Summary
* Adds a new `wp_irving_component_config_after_hydration` filter.
* Implements a new `interstitials` template key for the `irving/post-list` component.

Example,
```json
{
  "name": "irving/post-list",
  "config": {
    "templates": {
      "interstitials": {
        "3": [
          {
            "name": "your-interstitial-here/position-3"
          }
        ],
        "10": [
          {
            "name": "your-interstitial-here/position-10"
          }
        ]
      }
    }
  }
}
```

## Ticket
[IRV-617](https://alleyinteractive.atlassian.net/browse/IRV-617)

### DIFF
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -555,6 +555,17 @@ class Component implements JsonSerializable {
 			$new_config = call_user_func_array( $this->config_callback, [ $new_config ] );
 		}
 
+		/**
+		 * Filter the config values after the config callback fires.
+		 *
+		 * Note: This fitler uses `after_hydration` instead of `post_hydration`
+		 * to avoid confusion with `post` in the WordPress context.
+		 *
+		 * @param array  $config Config for this component.
+		 * @param string $name   Name of this component.
+		 */
+		$new_config = apply_filters( 'wp_irving_component_config_after_hydration', $new_config, $this->get_name() );
+
 		// @todo Add error handling.
 		if ( is_array( $new_config ) ) {
 			$this->set_config( $new_config );
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -558,7 +558,7 @@ class Component implements JsonSerializable {
 		/**
 		 * Filter the config values after the config callback fires.
 		 *
-		 * Note: This fitler uses `after_hydration` instead of `post_hydration`
+		 * Note: This filter uses `after_hydration` instead of `post_hydration`
 		 * to avoid confusion with `post` in the WordPress context.
 		 *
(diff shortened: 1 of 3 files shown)


inc/components/class-component.php
```diff
@@ -555,6 +555,17 @@ private function hydrate_config(): self {
 			$new_config = call_user_func_array( $this->config_callback, [ $new_config ] );
 		}
 
+		/**
+		 * Filter the config values after the config callback fires.
+		 *
+		 * Note: This fitler uses `after_hydration` instead of `post_hydration`
```
amyevans, line 561:
```suggestion
		 * Note: This filter uses `after_hydration` instead of `post_hydration`
```

inc/components/class-component.php
```diff
@@ -557,6 +557,17 @@ private function hydrate_config(): self {
...
+		/**
+		 * Filter the config values after the config callback fires.
+		 *
+		 * Note: This filter uses `after_hydration` instead of `post_hydration`
+		 * to avoid confusion with `post` in the WordPress context.
+		 *
+		 * @param array  $config Config for this component.
+		 * @param string $name   Name of this component.
+		 */
+		$new_config = apply_filters( 'wp_irving_component_config_after_hydration', $new_config, $this->get_name() );
```
joemcgill, line 569: Can we just shorten this one to `wp_irving_component_config` or is there a reason to retain that filter name for some other purpose? The shorter version would have parity with the `wp_irving_component_children` filter which fires after the `children_callback` method and child hydration is finished.

tests/components/test-components.php
```diff
@@ -442,14 +442,19 @@ public function test_component_post_list() {
...
 			],
-			'after'   => [
+			'after'         => [
 				[ 'name' => 'example/after' ],
 			],
-			'wrapper' => [ 'name' => 'example/wrapper' ],
-			'item'    => [ 'name' => 'example/item' ],
+			'wrapper'       => [ 'name' => 'example/wrapper' ],
+			'item'          => [ 'name' => 'example/item' ],
+			'interstitials' => [
```
joemcgill, line 453: I think we should make a separate test specifically for interstitials, which includes multiple positions.

inc/components/class-component.php
jameswburke: I think it's useful to clarify since the other one is `wp_irving_component_config_pre_hydration`?

joemcgill: Then should we rename the `wp_irving_component_children` filter in the same way?

PR #227, opened 2020-08-18
Implement irving/term-list component

jameswburke:
## Summary
Implements a new `irving/term-list` component, which exposes a [`WP_Term_Query`](https://developer.wordpress.org/reference/classes/wp_term_query/) object.

## Screenshot
Screen Shot 2020-08-18 at 3 33 56 PM

## Example
Display all the tags for a given post in `single.json`
```json
{
  "name": "irving/term-list",
  "config": {
    "taxonomy": "post_tag",
    "templates": {
      "item": [
        {
          "name": "irving/term-link",
          "children": [
            {
              "name": "irving/term-name"
            }
          ]
        }
      ]
    }
  },
  "use_context": {
    "irving/post_id": "object_ids"
  }
}
```

## Note for reviewer
There is a lot of code duplicated from the `irving/post-list` component. While a refactor/abstraction would be useful, it's outside the scope of this work for now.

### DIFF
--- a/inc/components/components/term-list/component.json
+++ b/inc/components/components/term-list/component.json
@@ -1,0 +1,26 @@
+{
+  "name": "irving/term-list",
+  "_alias": "irving/fragment",
+  "category": "Term.",
+  "config": {
+    "templates": {
+      "default": [],
+      "hidden": true,
+      "type": "array"
+    },
+    "object_ids": {
+    },
+    "query_args": {
+      "default": [],
+      "hidden": true,
+      "type": "array"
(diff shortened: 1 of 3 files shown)


inc/components/components/term-list/component.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "irving/term-list",
+  "_alias": "irving/fragment",
+  "category": "Term.",
```
joemcgill, line 4: Period seems harsh 😆

PR #232, opened 2020-08-24
Components: Add coral-comment-counts component

joemcgill:
This adds a new component named `irving/coral-comment-counts` that can be used
for displaying the number of comments on a specific post.

Other updates:

* Adds field on the integrations settings page for setting the Coral URL value.
* Adds a new helper function `WP-Irving\Integrations\get_option()` for getting a specific integration option value. See e9e7b94.

### DIFF
--- a/tests/components/test-components.php
+++ b/tests/components/test-components.php
@@ -1439,7 +1439,7 @@ class Test_Components extends WP_UnitTestCase {
 			[
 				'config' => [
 					'articleUrl' => get_the_permalink( self::$post_id ),
-					'embedUrl' => 'https://example.coral.test',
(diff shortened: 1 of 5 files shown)


tests/components/test-components.php
```diff
@@ -1426,6 +1426,39 @@ public function test_component_video() {
...
+					'embed_URL' => 'https://example.coral.test',
+					'post_id'   => self::$post_id,
+				],
+			]
+		);
+
+		$this->assertComponentEquals( $expected, $component );
+	}
+
+
```
joemcgill, line 1461:
```suggestion
```

PR #241, opened 2020-08-28
DEF-193: Allow user to set custom username through Pico/Coral integration

tylerreckart:


inc/integrations/class-coral.php
```diff
@@ -151,9 +151,12 @@ public function get_endpoint_settings(): array {
 	 * @param \WP_REST_Request $request The request object.
 	 */
 	public function process_validate_endpoint_request( \WP_REST_Request $request ) {
+		// Allow access from the frontend.
+		header( 'Access-Control-Allow-Origin: ' . home_url() );
```
tylerreckart, line 155: @joemcgill Without this being defined I'm running into issues with cors. I'm flagging this as something that needs to be updated, but I did not want this to block this round of work from being reviewable.

joemcgill: Thanks for confirming. If you're running the old version of the Pico plugin, that might be the case, since they were removing the native CORS headers from WP, but we should investigate. I'll open a ticket.

joemcgill (approved):
🐉

PR #244, opened 2020-08-31
Hydrate WordPress data for images

joemcgill:
This updates the config schema for the irving/image component to allow
an attachment ID to be passed in as a template config argument and will
let WordPress automatically add the following attributes:

* src
* width
* height
* alt
* srcset
* sizes

Adds unit tests.


inc/components/components/image/component.json
```diff
@@ -28,35 +28,46 @@
...
+      "description":"A source list string for delivering responsive images.",
+      "type": "string"
+    },
     "style": {
       "default": [],
       "type": "array"
     },
-    "photon_transformations": {
-      "description": "A set of Photon API transformations in the order they should be performed",
+    "query_args": {
```
joemcgill, line 69: We can use this more generically for photon args or for other DAM/CDN arguments, when necessary.

inc/components/components/image/component.php
```diff
@@ -12,4 +12,28 @@
 /**
  * Register the component.
  */
-register_component_from_config( __DIR__ . '/component' );
+register_component_from_config(
+	__DIR__ . '/component',
+	[
+		'config_callback' => function( array $config ): array {
+
+			// Bail early if passed a `src` URL.
```
joemcgill, line 20: If a `src` is passed, I think we should respect that URL and not try to do any additional WordPress hydration, unless we're wanting to try and sideload the image, which seems potentially undesirable.

joemcgill:
Unsure why the `srcset` functionality is failing in `trunk`. Possibly a core bug that I'll investigate, but shouldn't block this PR.

PR #246, opened 2020-08-31
Improvements to <head> management, and block library support.

jameswburke:
## Summary
Updates to the WP Irving plugin to better support parts of <head> management, block library, and Yoast integration.

## Notes for reviewers
This PR was developed is in conjunction with https://github.com/alleyinteractive/irving/pull/286

This version will break any version of Irving without the above PR.

## Changelog entries
* **Added**: Block library styles will automatically enqueue to Irving frontend.
* **Added**: Support for the customizer's custom css.
* **Changed**: Updated `irving/head` component to include a `context` prop.
* **Changed**: Updated `irving/post-content` component to include the same props as `irving/text`, and enable html and oembed by default via the JSON config, rather than toggling via the config callback.
* **Changed**: Update the Yoast integration to pass the schema iva <head> rather than handling separately with an integration.

## Ticket(s)
* https://alleyinteractive.atlassian.net/browse/IRV-406

### DIFF
--- a/inc/components/components/post-content/component.json
+++ b/inc/components/components/post-content/component.json
@@ -3,6 +3,28 @@
   "_alias": "irving/text",
   "category": "Post",
   "description": "Get the post content.",
+  "config": {
+    "content": {
+      "default": "",
+      "type": "string"
+    },
+    "html": {
+      "default": true,
+      "type": "bool"
+    },
+    "oembed": {
+      "default": true,
+      "type": "bool"
+    },
+    "style": {
+      "default": [],
+      "type": "array"
+    },
+    "tag": {
+      "default": "span",
+      "type": "string"
+    }
+  },
   "use_context": {
     "irving/post_id": "post_id"
   }
--- a/inc/components/components/post-content/component.json
+++ b/inc/components/components/post-content/component.json
@@ -21,7 +21,7 @@
       "type": "array"
     },
     "tag": {
-      "default": "span",
+      "default": "div",
       "type": "string"
     }
   },
--- a/inc/components/components/post-content/component.php
+++ b/inc/components/components/post-content/component.php
@@ -31,11 +31,7 @@ register_component_from_config(
 			 *
 			 * @see https://github.com/WordPress/gutenberg/blob/30cd85aebe14eee995e3162f09d31f4d4786f101/packages/block-library/src/post-content/index.php#L23
 			 */
-			$content = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', get_the_content( null, false, $post_id ) ) ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
-
-			$config['content'] = $content;
-			$config['html']    = true;
-			$config['oembed']  = true;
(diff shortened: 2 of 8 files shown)


inc/components/components/post-content/component.json
```diff
@@ -3,6 +3,28 @@
...
+    "oembed": {
+      "default": true,
+      "type": "bool"
+    },
+    "style": {
+      "default": [],
+      "type": "array"
+    },
+    "tag": {
+      "default": "span",
```
joemcgill, line 24: Why not a `div` for the default?

inc/components/components/post-content/component.php
```diff
@@ -31,11 +31,7 @@
 			 *
 			 * @see https://github.com/WordPress/gutenberg/blob/30cd85aebe14eee995e3162f09d31f4d4786f101/packages/block-library/src/post-content/index.php#L23
 			 */
-			$content = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', get_the_content( null, false, $post_id ) ) ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
-
-			$config['content'] = $content;
-			$config['html']    = true;
-			$config['oembed']  = true;
+			$config['content'] = apply_filters( 'the_content', str_replace( ']]>', ']]&gt;', get_the_content( null, false, $post_id ) ) ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
```
joemcgill, line 34: 💪 yeah, this is way better. Default config, FTW!

inc/templates/namespace.php
```diff
@@ -641,3 +643,88 @@ function inject_favicon( array $children, array $config, string $name ): array {
...
+
+/**
+ * Inject the custom css styles into the <head> component.
+ *
+ * @param array  $children Children for this component.
+ * @param array  $config   Config for this component.
+ * @param string $name     Name of this component.
+ * @return array
+ */
+function inject_custom_css( array $children, array $config, string $name ): array {
```
joemcgill, line 655: I think it might be cleaner to declare these functions in the `irving/head` component file, since it's a filter specifically for that component, or even to add an action inside the `children_callback` that these could hook to.

inc/components/components/post-content/component.json
jameswburke: The default for text is `span`, so I copied that over, but `div` makes sense to me too.

inc/templates/namespace.php
jameswburke: Do you think the `add_action` calls should still be in namespace? Or somewhere else?

I'm not against adding a component-specific filter, but based on convos we've had before, I thought we were trying to avoid that pattern in favor of doing it via the `class-component.php`.

jameswburke:
> I'm curious what causes this to break anything running current and past versions of the `@irvingjs` code? Anything we can do to maintain backwards compatibility?
This will break the `irving/head` component in `@irvingjs/styled-components` unless the companion in https://github.com/alleyinteractive/irving/pull/286 is used.

The only thing I can think of would require passing the various Irving package versions back to WP Irving via the components endpoint.

> I'm also curious if there is a more generic way that we can grab any scripts/styles that have been enqueued and inject them into our helmet container rather than needing to explicitly handle each one by injecting them via filter or as children via `Components\html_to_components()`.
We have a spike for this. Until we've walked through it and make a decision as a group, I prefer cherry-picking the functionality we want to avoid breaking changes in the future, and enqueuing things we don't want/need. A lot of these plugins are adding scripts or styles that aren't applicable to the Irving frontend. I'm hesitant to inject them carte-blanche.

PR #247, opened 2020-09-01
irving/post-list item override support

jameswburke:
Adds support for a new `item_overrides` template option on `irving/post-list`. Similar to the interstitial syntax, it allows the post list to define an alternative item for a specific position.

For example, you can display a river of cards, and the first one to be a hero, more prominently highlighting the content.

```json
{
  "item": [
    {
      "name": "river-card"
    }
  ],
  "item_overrides": {
    "0": [
      {
        "name": "hero-card"
      }
    ]
  }
}
```


inc/components/components/post-list/component.php
```diff
@@ -62,15 +63,23 @@
...
+					$item = $templates['item_overrides'][ $index ] ?? $templates['item'];
+
+					// Ensure single items are wrapped in an array.
+					$item = ( isset( $item[0] ) ) ? $item : [ $item ];
+
 					return [
 						'name'     => 'irving/post-provider',
 						'config'   => [
 							'post_id' => $post_id,
+							'index'   => $index++,
```
joemcgill, line 82: Is this going to make the index off by one compared to the position in the array? I would think that you would want to create the post-provider with the current index.

You could also do something like `array_map( callback_function( $item, $index ) {...}, $array, array_keys( $array ) )` to pass the index as the second param of the callback function. That said, a `foreach` would also be fine.

jameswburke: Since it's a post increment operator, it returns $index, then adds one for the next loop around.

I'll look at the array keys option again though, I wasn't having luck earlier.

PR #249, opened 2020-09-01
Implement local WP username storage for Coral users

xmattus:


inc/integrations/class-coral.php
```diff
@@ -117,40 +150,109 @@ public function get_endpoint_settings(): array {
...
-			'jti'  => uniqid(),
-			'exp'  => time() + ( 90 * DAY_IN_SECONDS ), // JWT will expire in 90 days.
-			'iat'  => time(),
-			'user' => [
-				'id'       => $verified_user['id'],
-				'email'    => $verified_user['email'],
-				'username' => $verified_user['username'],
-			],
+		// Check for existing username, since the verification call returned 200 OK.
+		$username = $this->get_username( $verified_data['user']['email'] );
```
joemcgill, line 171: If the `$user_obj` isn't filtered by `wp_irving_verify_coral_user`, the `user` key will throw an undefined index error. We could either modify the filter call so that it always returns `false` and passes the `$user_obj` as a second param or write a sanity check here that returns a failed status if we can't confirm a username.

inc/integrations/class-coral.php
```diff
@@ -193,4 +295,157 @@ public function build_jwt( array $credentials ): string {
 	public function base64url_encode( string $data ): string {
 		return rtrim( strtr( base64_encode( $data ), '+/', '-_' ), '=' );
 	}
+
+
```
joemcgill, line 299: Remove extra line.

inc/integrations/class-coral.php
```diff
@@ -193,4 +295,157 @@ public function build_jwt( array $credentials ): string {
...
+	/**
+	 * Retrieve a username record. These functions use the post_title column
+	 * for the email address and the post_excerpt column for the username.
+	 *
+	 * @param string $email    The email address of the user.
```
joemcgill, line 304: Unnecessary spaces after the `$email` argument in this docblock.

inc/integrations/class-coral.php
```diff
@@ -117,40 +150,109 @@ public function get_endpoint_settings(): array {
...
+
+	/**
+	 * Accept requests to set a username for a given email address.
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 * @return array|\WP_REST_Response
+	 */
+	public function process_set_username_endpoint_request( \WP_REST_Request $request ) {
+		// Allow access from the frontend.
+		header( 'Access-Control-Allow-Origin: ' . home_url() );
```
joemcgill, line 208: I believe the REST API adds this by default and should be unnecessary. If it _is_ needed, than we should probably be filtering the `http_origin` filter and let the `rest_send_cors_headers()` function handle this.

inc/integrations/class-coral.php
```diff
@@ -193,4 +295,157 @@ public function build_jwt( array $credentials ): string {
...
+	}
+
+	/**
+	 * Retrieve a username record's post ID. These functions use the post_title column
+	 * for the email address and the post_excerpt column for the username.
+	 *
+	 * @param string $email    The email address of the user.
+	 * @return int The post ID, or 0 if none is found.
+	 */
+	private function get_username_post_id( $email ) : int {
```
joemcgill, line 334: I'm not sure we should be using the email address as the relevant information to match with a Pico user, since you can update your email address in Pico. Instead, could we store the Pico user ID and use that to lookup usernames?

inc/integrations/class-coral.php
```diff
@@ -193,4 +295,157 @@ public function build_jwt( array $credentials ): string {
...
+		if ( ! $stored_hash || ! $hash || $hash !== $stored_hash ) {
+			return false;
+		}
+
+		// Stop if the username is already taken by someone else.
+		if ( $this->username_exists( $username ) ) {
+			return false;
+		}
+
+		$args = [
```
joemcgill, line 408: Same concern here as above about using the email address as the primary identifier linking our user record to a Pico user. Can we use the Pico ID instead?

inc/integrations/class-integrations-manager.php
```diff
@@ -74,6 +74,13 @@ public function group_and_format_options_for_storage( array $options ): array {
 					// the components JSON endpoint.
 					$formatted_options['coral']['private'] = true;
 					break;
+				// Build the config array for Pico.
+				case strpos( $key, 'pico_' ) !== false:
```
joemcgill, line 78: 👍

inc/integrations/class-coral.php
tylerreckart: 👍  To @joemcgill's suggestion. The ID's _should_ be globally unique in Pico and if a user changes their email address that value should persist.

tylerreckart: The email definitely works as the md5 reference, though.

inc/integrations/class-coral.php
```diff
@@ -162,13 +159,18 @@ public function process_validate_endpoint_request( \WP_REST_Request $request ) {
 		// Verify the user's credentials.
 		$verified_data = apply_filters( 'wp_irving_verify_coral_user', $user_obj );
 
-		// Bail early if the verified user doesn't exist.
-		if ( empty( $verified_data ) || ! $verified_data ) {
+		// Bail early if the verified user isn't confirmed by the verification filter.
+		if (
+			empty( $verified_data ) ||
+			! $verified_data ||
```
joemcgill, line 165: `empty()` evaluates `false` as empty, so I think this line is unnecessary. See: https://www.php.net/manual/en/function.empty.php

PR #254, opened 2020-09-03
DEF-235: Add allowed tiers input for Pico/Coral SSO to Irving Integrations settings page

tylerreckart:
Adds a new input where users can add comma-separated values to pass whitelisted tiers to the `<PicoObserver />` component on the front-end.

### DIFF
--- a/inc/integrations/class-pico.php
+++ b/inc/integrations/class-pico.php
@@ -111,7 +111,7 @@ class Pico {
 		?>
 			<input id="pico_tiers" type="text" name="irving_integrations[<?php echo esc_attr( 'pico_tiers' ); ?>]" value="<?php echo esc_attr( $tiers ); ?>" />
 			<label for="pico_tiers">
-				<em><?php echo esc_html__('Tiers should be input as comma-separated values (e.g. Reader,Subscriber)', 'wp-irving') ?></em>
+				<em><?php echo esc_html__( 'Tiers should be input as comma-separated values (e.g. Reader,Subscriber)', 'wp-irving' ); ?></em>
 			</label>
 		<?php
 	}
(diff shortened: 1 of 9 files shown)


inc/integrations/class-pico.php
```diff
@@ -77,6 +78,15 @@ public function register_settings_fields() {
 			'wp_irving_integrations',
 			'irving_integrations_settings'
 		);
+
+		// Register new fields for the Coral integration.
+		add_settings_field(
+			'wp_irving_pico_tiers',
+			esc_html__( 'Pico/Coral Whitelisted SSO Tiers', 'wp-irving' ),
```
joemcgill, line 85: Avoid "Whitelist" in favor of more inclusive language, here and throughout 🔢 

```suggestion
			esc_html__( 'Allowed Pico/Coral SSO Tiers', 'wp-irving' ),
```

inc/integrations/class-pico.php
```diff
@@ -91,6 +101,19 @@ public function render_pico_auth_key_input() {
...
 
+	/**
+	 * Render an input for the Pico Whitelisted SSO Tiers.
+	 */
+	public function render_pico_tiers_input() {
+		// Check to see if there are existing tiers in the option.
+		$tiers = $this->options[ $this->option_key ]['tiers'] ?? '';
+
+		?>
+			<input type="text" name="irving_integrations[<?php echo esc_attr( 'pico_tiers' ); ?>]" value="<?php echo esc_attr( $tiers ); ?>" />
```
joemcgill, line 112: We may want to add some help text that explains that this needs to be a comma separated list.

joemcgill (approved):
Couple of comments, but pre-approving pending updates. 🌵

inc/integrations/class-pico.php
tylerreckart: Great point! I hadn't even considered the terms lack of inclusivity.

PR #255, opened 2020-09-04
Add initial data layer to GTM integrations.

joemcgill:
Adds: WP_Irving\Integrations\Google_Tag_Manager\get_data_layer() that injects a data layer object into all page responses.
Adds: Filter `wp_irving_gtm_data` to more easily allow filtering the GTM data.

### DIFF
--- a/inc/integrations/class-google-tag-manager.php
+++ b/inc/integrations/class-google-tag-manager.php
@@ -30,7 +30,7 @@ class Google_Tag_Manager {
 	private $options;
 
 	/**
-	 * Setup the singleton. Validate JWT is installed, and setup hooks.
+	 * Setup the singleton.
 	 */
 	public function setup() {
 		// Retrieve any existing integrations options.
@@ -39,7 +39,7 @@ class Google_Tag_Manager {
 		// Register settings fields for integrations.
 		add_action( 'admin_init', [ $this, 'register_settings_fields' ] );
 
-		// Filter the integrations manager to include our Pico props.
+		// Filter the integrations manager to include a data layer for GTM.
 		add_filter( 'wp_irving_integrations_option', [ $this, 'get_data_layer' ] );
 	}
 
@@ -76,11 +76,10 @@ class Google_Tag_Manager {
 	 * @return array An array of options.
 	 */
 	public function get_data_layer( $options ) {
-
 		$data_layer = apply_filters(
 			'wp_irving_gtm_data',
 			[
-				'title' => wp_title( null, false ),
+				'event' => 'irving.page',
 			]
 		);
 
--- a/inc/integrations/class-google-tag-manager.php
+++ b/inc/integrations/class-google-tag-manager.php
@@ -38,6 +38,9 @@ class Google_Tag_Manager {
 
 		// Register settings fields for integrations.
 		add_action( 'admin_init', [ $this, 'register_settings_fields' ] );
+
+		// Filter the integrations manager to include our Pico props.
+		add_filter( 'wp_irving_integrations_option', [ $this, 'get_data_layer' ] );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


joemcgill:
@davisshaver I've made a couple of adjustments in 399ea7e. I've removed the `event` key, because the GTM event name is handled in the IrvingJS GTM integration [here on all renders](https://github.com/alleyinteractive/irving/blob/master/packages/integrations/components/googleTagManager/index.js#L41) and [here on initial renders](https://github.com/alleyinteractive/irving/blob/master/packages/integrations/components/googleTagManager/index.js#L58) and data passed from the server would override that event value. If you want to pass your own event name, you can do that in a filter, but should know that on initial render, you will likely get double events.

I'm also passing two default values: the page title and the URL, since those are commonly used. We can expand on these, but this should be ok for launch + anything you are adding with the filter.

You can test that this is working with the fixes to the IntegrationsManger in https://github.com/alleyinteractive/irving/pull/303.

inc/integrations/class-google-tag-manager.php
```diff
@@ -65,4 +70,29 @@ public function render_container_id_input() {
...
+		/**
+		 * Filters the data layer provided to GTM for each page.
+		 *
+		 * @param array $data_layer The data layer arguments for GTM.
+		 */
+		$data_layer = apply_filters(
+			'wp_irving_gtm_data',
+			[
+				'title' => wp_title( null, false ),
+				'url'   => get_permalink() ?: home_url(),
```
davisshaver, line 90: I think this would only cover `is_single()` and `is_home()` queries for URL purposes, should we hold this until we have a drop-in function that can handle all cases?

inc/integrations/class-google-tag-manager.php
```diff
@@ -65,4 +70,29 @@ public function render_container_id_input() {
...
+	public function get_data_layer( $options ) {
+		/**
+		 * Filters the data layer provided to GTM for each page.
+		 *
+		 * @param array $data_layer The data layer arguments for GTM.
+		 */
+		$data_layer = apply_filters(
+			'wp_irving_gtm_data',
+			[
+				'title' => wp_title( null, false ),
```
davisshaver, line 89: I think this is slightly redundant with `document.title` but no real argument from me against passing it. I am wondering though if these should be namespaced under `irving`, as `title` would be a very common top-level data layer key.

davisshaver (approved):
@joemcgill  good point about double events/overwriting `event`. I think this looks just about ready, one question about namespacing for built-in data layer properties and then 🚢

inc/integrations/class-google-tag-manager.php
joemcgill: You're right that it's redundant, but I've seen occasions where the event fires before document.title is updated on the front end, so this ensures that we're always getting the right title from WP.

inc/integrations/class-google-tag-manager.php
joemcgill: Good catch on this. I think we can just reuse the path passed to the REST API endpoint rather than  inferring the URL from the query. Fixed in 9b6c950

joemcgill:
@davisshaver Good idea about namespacing this data inside a custom `irving` property. Handled in 7288ae3. I also updated the way URLs are generated for this response. Can I get another look at these before I commit?

inc/templates/namespace.php
```diff
@@ -552,16 +553,21 @@ function setup_integrations( array $data, WP_Query $query, string $context ): ar
...
-	 * Modify the value stored in the `irving_integraionts` option before it's
+	 * Modify the value stored in the `irving_integrations` option before it's
 	 * used.
 	 *
 	 * This is essentially an alias for the `option_{$option}` filter, but
 	 * by declaring this explicitly, we can make it more clear to developers
 	 * how to modify the value correctly.
 	 *
-	 * @var array
+	 * @param array $options Array of integration options.
```
davisshaver, line 563: spacing off on first param docs?

davisshaver (approved):
🏷️ 🕴️

PR #258, opened 2020-09-09
add fragment property to post permalink component config

xmattus:


davisshaver (approved):
#️⃣

jameswburke (commented):
I think this would be more useful on the `irving/link` component, which permalink extends (and thus would still have access to).

Additionally, the url set may already have a fragment, so I'd recommend some additional logic that considers that edge case.

PR #261, opened 2020-09-15
Bug/def 324/add checkbox for pico staging urls

jameswburke:
## Summary
Pico is rolling out staging accounts, but these require using a different API url.

This PR adds a checkbox which toggles the API url as needed.

## Notes for reviewers
None.

## Changelog entries
* **Added**: Support for using Pico staging accounts.

## Ticket(s)
* https://alleyinteractive.atlassian.net/browse/DEF-324

### DIFF
--- a/inc/integrations/class-pico.php
+++ b/inc/integrations/class-pico.php
@@ -64,6 +64,17 @@ class Pico {
 		}
 
 		add_filter( 'wp_irving_verify_coral_user', [ $this, 'verify_pico_user_for_sso' ] );
+
+		// Use staging API urls.
+		if ( get_option( 'irving_integrations' )['pico']['use_staging'] ?? false ) {
+			if ( ! defined( 'PP_API_ENDPOINT' ) ) {
+				define( 'PP_API_ENDPOINT', 'https://api.staging.pico.tools' );
+			}
+
+			if ( ! defined( 'PP_WIDGET_ENDPOINT' ) ) {
+				define( 'PP_WIDGET_ENDPOINT', 'https://widget.staging.pico.tools' );
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


inc/integrations/class-pico.php
```diff
@@ -64,6 +64,17 @@ public function setup() {
 		}
 
 		add_filter( 'wp_irving_verify_coral_user', [ $this, 'verify_pico_user_for_sso' ] );
+
+		// Use staging API urls.
+		if ( get_option( 'irving_integrations' )['pico']['use_staging'] ?? false ) {
```
mslinnea, line 69: If `irving_integrations` option doesn't exist, this will throw a PHP warning because of trying to access array indexes in something that is not an array.  `get_option` takes a default as a second param... setting the default to an array should solve this.

inc/integrations/class-pico.php
```diff
@@ -64,6 +64,17 @@ public function setup() {
...
+			if ( ! defined( 'PP_API_ENDPOINT' ) ) {
+				define( 'PP_API_ENDPOINT', 'https://api.staging.pico.tools' ); // phpcs:ignore
```
mslinnea, line 71: What phpcs rule is being ignored here?  I don't see an issue...

mslinnea (approved):
🐴

inc/integrations/class-pico.php
jameswburke: I confirmed that the null coalescing doesn't mind,

Screen Shot 2020-09-15 at 6 23 20 PM

inc/integrations/class-pico.php
jameswburke: It wants `PP_API_ENDPOINT` to have a WP Irving namespace.

PR #262, opened 2020-09-16
Add class_name and style as default configs

amyevans:
Ref: [IRV-606](https://alleyinteractive.atlassian.net/browse/IRV-606)

This PR:
* Adds `class_name` as a default prop for all components. It accepts an array of values that is output as a string (which is what [Irving Core expects](https://github.com/alleyinteractive/irving/blob/master/packages/styled/hooks/useStandardProps.js)).
* Adds `style` as a default prop for all components. It is an associative array of style properties => values.
* Updates tests and existing component defaults to account for the changes.
* Note that `theme_name` is already a default prop so no further changes were required to meet that A/C.


jameswburke (approved):
:bee:

PR #263, opened 2020-09-16
Pico: Add support for using the staging widget URL

joemcgill:
This passes the widget endpoint URL as part of the integration response so that our front end application can swap out the location for the widget URL when switched into staging mode.


jameswburke (approved):
:bee:

PR #265, opened 2020-09-18
Add options for camel casing keys

amyevans:
## Summary
Adds two new options for the schema of a config, `camel_case_config_key` and `camel_case_inner_keys`.

* If not explicitly set, then they are assumed `true` and all keys and inner keys get camel cased.
* If `camel_case_config_key` is set to `false`, that config's key is not transformed to camel case.
* If `camel_case_inner_keys` is set to `false` and the config is an object, the inner keys are not transformed to camel case.

## Ticket
[IRV-707](https://alleyinteractive.atlassian.net/browse/IRV-707)

### DIFF
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -921,12 +921,20 @@ class Component implements JsonSerializable {
 		// Loop through each key.
 		foreach ( $array as $key => $value ) {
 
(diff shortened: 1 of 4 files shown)


jameswburke (approved):
:bee:

PR #267, opened 2020-09-28
Fix setting global WordPress properties

joemcgill:
This removes some manual overrides in the WP_Irving\REST_API\Components_Endpoint
class in favor of directly using methods from core WordPress for setting
global values of the $wp_query object and other globals that are set during
the main WP loading process.

This also fixes a bug where homepage paths would not allow the $wp_query->is_home
flag to be properly set, even after the fixes in WordPress 5.5 were made
see: https://core.trac.wordpress.org/changeset/48053


jameswburke (approved):
:bee:

PR #269, opened 2020-09-30
GTM: Escape the site path in the data layer.

joemcgill:
This patches a potential security issue if an XSS payload is added to the URL string.


jameswburke (approved):
:bee:

PR #271, opened 2020-10-01
Implement site-theme-provider and template-part components

jameswburke:
## Summary
As titled.

## Notes for reviewers
None.

## Changelog entries
* **Added**: `irving/site-theme-provider` component, tests, and [documentation](http://storybook-qa.irvingjs.com/?path=/docs/wordpress-components-site-theme-provider--page).
* **Added**: `irving/template-part` component, tests, and [documentation](http://storybook-qa.irvingjs.com/?path=/docs/wordpress-components-template-part--page).

## Ticket(s)
None.


jomurgel (approved):
Looks good! 🍞

joemcgill:
Thanks @jameswburke! I think I understand what problems you're trying to solve with these components, but just to be certain, I'll describe my assumptions.

**For the template part component:**

It seems like you're looking for a way to dynamically load template parts based on a "slug" value, which I assume would work similar to the traditional `get_template_part()` function in WordPress themes, which I could see being really handy. I'm curious if we could consolidate the JSON schema with the way we are currently describing template parts so people aren't confused by the difference between these two forms:

```json
{
	"name": "template-parts/post-card"
}

{
	"name": "irving/template-parts"
	"slug": "post-card"
}
```

I could see us going a few directions here. The WordPress equivalent lets you choose variations on a template name by passing a second parameter, so if you have two different headers (for example) we could support something like:

```json
// Load the default header in template-parts/header.json
{
	"name": "template-parts/header"
}

// Load an alternate header in template-parts/header-slim.json if available and fall back to template-parts/header.json if not.
{
	"name": "template-parts/header"
	"type": "slim"
}
//
```

This would work for many use cases, but not for a use case where you wanted to conditionally choose between two very different template types (e.g., a post card and a video promotion, for example). For that use case we would need to rely on something more like a conditional helper component where you would do something like this (pseudo-code):

{
	"name": "irving/conditional-loader"
	"config": {
 		"options": [
			{
				"case": "video",
				"name": "irving/video-promo"
				"config": {...},
				"children": {...}
			},
			{
				"case": "podcast",
				"name": "irving/podcast-promo"
				"config": {...},
				"children": {...}
			},
			{
				"case": "default",
				"name": "irving/post-card"
				"config": {...},
				"children": {...}
			},
		]	
	},
	"use_context": {
		"defector/homepage-promo": "case"
	}
}

Also, I assume you're trying to solve the current limitation where template parts aren't being hydrated when added to a component via the `wp_irving_component_children` filter, which I agree is something to be solved and is something we could handle regardless of which convention we choose for our JSON templating data structure.

**For site theme context**

I assume you're looking for ways to pass site-theme data into a subset of components. Using the context API for this is exactly the direction I think we should go. I would love for this to work like a React developer would expect a `<ThemeProvider>` component to work ([see docs](https://styled-components.com/docs/advanced#theming)) where we pass the default theme properties as context in the root of the page (like we do for the global `$wp_query` and `$post` objects) and let the theme provider override specific values to a subset of children below.

jameswburke:
I'm okay merging this with core templating functionality. Also agree that we need to solve the template part hydration, but I don't currently have a solution for either of those.

If we want to follow WP's lead on `get_template_part()` I'd prefer to see it implemented so you can provide both the slug and name,
```json
{
	"name": "irving/template-part",
	"config": {
		"slug": "header",
		"name": "default"
	}
}
```

which would work the same as,
```json
{
	"name": "template-part/header-default"
}```

Exposing those parameters via configs so that the context provider can modify them is a big part of the problem I'm trying to solve with these solutions.

--------------------

I considered using context for the Site Theme Provider, but we'd need to introduce some way of traversing the tree to get our specific value. IMO we're heavily limited by JSON here,

```json
{
	"name": "irving/text",
	"use_context": {
		"irving/site-theme.header.templatePart": "content"
	}
}
```

It would be similar for a `irving/get_option` provider, or if you attempted to use an array/object from post meta.

Additionally, if we expose the entire site theme to the context provider, we have to compute the entire tree first. Whereas using `get_site_theme()` as I've written only computes the final value being asked for. Probably not a huge performance concern, but it does exist.

joemcgill:
> If we want to follow WP's lead on get_template_part() I'd prefer to see it implemented so you can provide both the slug and name...

I agree, I think that would be a nice way to handle variable template parts, though I agree that it's not completely clear on how we would recommend someone passing context in that scenario without adding several new default context values or coming up with a way of referencing specific properties of a context value.

We could rely on providers like our post-provider to enable more than just the post ID to be passed as context to the template, like this:

```json
// post-provider/component.json
{
  "name": "irving/post-provider",
  "_alias": "irving/fragment",
  "category": "Data",
  "description": "Post context provider.",
  "config": {
    "post_id": {
      "default": 0,
      "type": "integer"
    },
    "post_type": {
      "type": "post"	  
      "type": "string"
    }
  },
  "provides_context": {
    "irving/post_id": "post_id"
    "irving/post_type": "post_type"
  }
}

// template file .json
{
  "name": "irving/post-provider",
  "children": [
    {
      "name": "irving/template-part",
      "config": {
        "slug": "header",
        "name": "default"
      },
      "use_context": {
        "irving/post_type" : "name"
      }
    }
  ]
}
```

Consuming specific properties of a context value is much more difficult, I think. We could introduce a more verbose syntax for `use_context` perhaps, which would allow you to identify both the object you want to consume from context and the values you want to map to your component. Something like this maybe?

```
{
  "name": "irving/post-card",
  "use_context": [
    { 
      "key" : "irving/settings",
      "map" : {
        "background-color": "color.primary",
        "show-eyebrow": "post-card.show-eyebrow",
        "thumbnail-crop": "post-card.thumbnail-crop"
      }
    }
  ]
}
```

That way we could consume a global settings object and allow sub-components to update particular settings by updating context.

jameswburke:
To your first point, we could expose the entire WP Post object via the provider, and use the more verbose `use_context` syntax as you outlined.

It's funny you mention that too, Owen, Tyler, and I actually landed on the same idea during our Friday discussion. Our syntax looked more like this,

```json
{
  "name": "irving/post-card",
  "use_context": {
    "irving/settings": {
      "background-color": "color.primary",
      "show-eyebrow": "post-card.show-eyebrow",
      "thumbnail-crop": "post-card.thumbnail-crop"
    }
  ]
}
```

But I have no strong preference, yours does feel more readable and friendly.

joemcgill:
Oh nice! I think I like the syntax you all came up with better because it keeps the same object shape for each context key and gives us two different value shapes to support, 1) a string – the component's config value, 2) an object – spread context properties to config values. Should be easier to implement while keeping backwards compatibility too.

joemcgill:
@jameswburke is this PR still applicable, or has it been superseded by other site-theme and template-part work?

jameswburke:
Closing. Parts of the site theme provider could still be relevant, but we don't have clear direction currently, so not worth keeping alive.

PR #272, opened 2020-10-04
Feature/implement irving is main query

jameswburke:


inc/endpoints/class-components-endpoint.php
```diff
@@ -349,6 +359,7 @@ public function build_query() {
 		// This ensures we always parse the query so is_home and other flags
 		// get properly set before we handle template.
 		if ( empty( $query ) ) {
+			$wp_query->query( $query );
```
jameswburke, line 362: @joemcgill, I think we should consider adding this call to `query()`.

If `$query` is empty [in the constructor](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L3558), then [query()](https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-query.php#L3444) never runs, which means [get_posts()](https://github.com/WordPress/WordPress/blob/dc7bda962015e4be299a5a61c00a8d1c1c5c8b33/wp-includes/class-wp-query.php#L1763) doesn't run, which means the [`pre_get_posts` hook](https://github.com/WordPress/WordPress/blob/dc7bda962015e4be299a5a61c00a8d1c1c5c8b33/wp-includes/class-wp-query.php#L1779) never runs.

Which means, `pre_get_posts` never runs on the index/home/front-page query, because `$query` is an empty string.

Looking for a gut-check from you if this makes sense, or there's something I'm not seeing here.

PR #273, opened 2020-10-04
Add support for body classes

jameswburke:
## Summary
As titled.

## Notes for reviewers
Requires the `irving/body-wrapper`'s https://github.com/alleyinteractive/irving/pull/369

## Changelog entries
* **Added**: Automatically inject the body classes into `irving/head` component.

## Ticket(s)
None.

### DIFF
--- a/inc/templates/namespace.php
+++ b/inc/templates/namespace.php
@@ -21,6 +21,7 @@ function bootstrap() {
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_favicon', 10, 3 );
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_custom_css', 10, 3 );
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_block_library_styles', 10, 3 );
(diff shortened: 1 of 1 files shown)


ostowe (approved):
💯

inc/templates/namespace.php
```diff
@@ -21,6 +21,7 @@ function bootstrap() {
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_favicon', 10, 3 );
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_custom_css', 10, 3 );
 	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_block_library_styles', 10, 3 );
+	add_action( 'wp_irving_component_children', __NAMESPACE__ . '\\inject_body_classes', 10, 3 );
```
joemcgill, line 24: Why not handle this directly inside the children callback for the irving/head component?

jameswburke: Following the existing pattern, I'm fine with a refactor whenever. Could make an argument that having a filter gives devs an opportunity to unfilter it, but I don't feel super strongly about it.

PR #276, opened 2020-10-06
LEDE-329: Restructure and expand default analytics objects for WP Irving components

davisshaver:
Restructures and expands default analytics objects for WP Irving components.

Pairs with https://github.com/alleyinteractive/irving/pull/372.


juliobranha (approved):
🐐

PR #279, opened 2020-10-09
Limit paywall to posts by default

jameswburke:
## Summary
As titled.

## Notes for reviewers
None.

## Changelog entries
* **Added**: `wp_irving_pico_paywall_post_types` filter to determine which post types trigger the paywall.
* **Added**: `wp_irving_pico_options` filter for all the Pico options in the integrations manager.
* **Changed**: Pico will only filter the `post` post type by default.

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/IRV-776


amyevans (approved):
✨

PR #285, opened 2020-10-14
Update Alley Coding standards

alleyinteractive-ops:
This updates the project to make use of the Alley Interactive coding standards that are published to Packagist and updates our phpcs.xml file to reference the new "Alley-Interactive" sniff name.


tylerreckart (approved):
🧑‍🚀

PR #287, opened 2020-10-20
Modify caching in Coral integration

amyevans:
## Summary
This PR:
* Bypasses caching on the `/irving/v1/data/validate_sso_user` endpoint. (This helps resolve a bug where users were encountering a second Set Username modal after setting their Coral username for the first time, as the endpoint was returning stale data about the user).
* Adds caching on database queries in the Coral integration class
* Modifies minor code formatting

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/IRV-824

### DIFF
--- a/inc/integrations/class-coral.php
+++ b/inc/integrations/class-coral.php
@@ -68,6 +68,8 @@ class Coral {
 				}
 			);
 		}
+
+		add_action( "save_post_{$this->post_type}", [ $this, 'delete_cached_values' ], 10, 2 );
 	}
 
 	/**
@@ -523,10 +525,6 @@ class Coral {
 
 		$post = wp_insert_post( $args );
 
-		// Delete stored data for this SSO ID/username.
-		wp_cache_delete( "get_username_{$id}", $this->cache_group );
-		wp_cache_delete( "username_exists_{$username}", $this->cache_group );
-
 		return ( ! is_wp_error( $post ) );
 	}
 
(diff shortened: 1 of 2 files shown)


inc/integrations/class-coral.php
```diff
@@ -349,25 +350,38 @@ public function base64url_encode( string $data ): string {
...
+				LIMIT 1",
 				[
 					$id,
 					$this->post_type,
 				]
 			)
 		);
 
+		// Store the username in a transient.
+		// The transient is deleted when the username is updated.
```
davisshaver, line 380:
```suggestion
		/*
		 * Store the username in a transient.
		 * The transient is deleted when the username is updated.
		 */
```

inc/integrations/class-coral.php
```diff
@@ -378,25 +392,38 @@ private function get_username( $id ) : string {
...
+				LIMIT 1",
 				[
 					$id,
 					$this->post_type,
 				]
 			)
 		);
 
+		// Store the post ID in a transient.
+		// This can be set forever since the SSO ID to post ID relationship will never change.
```
davisshaver, line 422:
```suggestion
		/*
		 * Store the post ID in a transient.
		 * This can be set forever since the SSO ID to post ID relationship will never change.
		 */
```

inc/integrations/class-coral.php
```diff
@@ -407,29 +434,40 @@ private function get_username_post_id( $id ) : int {
...
+				LIMIT 1",
 				[
 					$username,
 					$this->post_type,
 				]
 			)
 		) ?? 0;
 
+		// Store the ID for the username in a transient.
+		// The transient is deleted when the username is added/updated.
```
davisshaver, line 468:
```suggestion
		/*
		 * Store the ID for the username in a transient.
		 * The transient is deleted when the username is added/updated.
		 */
```

davisshaver (approved):
🌽

inc/integrations/class-coral.php
```diff
@@ -472,6 +510,10 @@ private function set_username( $id, $username, $hash ) : bool {
 
 		$post = wp_insert_post( $args );
 
+		// Delete stored data for this SSO ID/username.
+		delete_transient( "get_username_{$id}" );
+		delete_transient( "username_exists_{$username}" );
```
joemcgill, line 515: These are great, though they won't account for a case where these Post values were updated by some other means. Probably not a problem at all, but sometimes it's better to do these cache clearing calls as a callback on the `update_post` or similar hook.

joemcgill (approved):
Not a blocker, but I'm thinking that many of these queries are fast enough that saving as transients might be unnecessary. Wrapping these in simple `wp_cache_set()` calls might be sufficient so they get pulled from a persistent object cache, when available, and then falling back to the DB queries might even be faster than the transient lookup.

inc/integrations/class-coral.php
amyevans: Cool, I moved them to `save_post_{$this->post_type}`

PR #289, opened 2020-10-21
IRV-682: Default components endpoint response to homepage when path is empty

tylerreckart:


joemcgill (approved):
Looks good to me. + 💯 for including tests for this!

joemcgill:
Once CS fixes are addressed, this should be good to merge.

PR #290, opened 2020-10-27
HOTFIX: WPCOM Legacy Redirector

davisshaver:
`wp_irving_components_redirect` filter is passed two arrays, second one being the `params`. No need to access through `get_params()`.


joemcgill (approved):
👍 Good catch! You mind updating the inline docs for the filter in `inc/endpoints/class-components-endpoint.php` while you've got this open?

PR #294, opened 2020-11-13
0.6.0 Release Prep

joemcgill:
- Bump version number
- Add Changelogs

Note: This removes the version number constant from the plugin
and the version number in composer.json, since both are unnecessary.
The release process docs are updated to reflect this change.


amyevans (approved):
✨

PR #296, opened 2020-11-24
support setting the theme in the component config object

xmattus:


inc/components/class-component.php
```diff
@@ -190,6 +190,11 @@ private function process_args( array $args ): self {
 			]
 		);
 
+		if ( ! empty( $args['config']['theme'] ) && 'irving/site-theme' !== $this->get_name() ) {
```
joemcgill, line 193: I think this is doing the opposite of what we want. If I'm reading this correctly, it's taking any `theme` set inside the `config` and moving it to a top level property, rather than taking any top-level `theme` property and moving it to `config`.

The latter, is what we want so that the theme property will be accessible inside the `config_callback` and `children_callback` functions as part of the `config` object.

xmattus: The precedence could go either way, but I took it to mean enabling the ability to pass a `theme` key in the config object (which would also expose it in the callbacks). I take your point that there's a lot of existing templates that use the top-level property, so maybe the order of precedence should run the other way?

tests/components/test-class-component.php
```diff
@@ -512,7 +521,7 @@ public function test_set_component_theme() {
 			]
 		);
 
-		$this->assertEquals( $expected, $legacy_component->to_array()['config'] );
```
joemcgill: These changes make me pause. The reason we were testing the output after running the `to_array()` method is because that method does some final tweaking too our config shape before returning a serialized JSON response. If the tests aren't passing without this change, then we probably still have a bug in our serialization step.

One thing that I think of is that during this method, the `theme_name` config property gets set from the top-level `theme` property, i.e., `$this->get_theme()`. If someone has modified the theme config property in a callback, this wouldn't be reflected in the output and we'll likely end up with a config object that looks something like this:

```js
"config": {
	"context": "site",
	"className": "",
	"style": [],
	"theme": "changedValue" // Altered in config.
	"themeName": "default", // Created during the to_array() method. Used by Irving.
	"themeOptions": [
		"default"
	]
},
```

joemcgill (approved):
Latest changes look good to me once PHPCS is 🟢.

PR #299, opened 2020-12-03
Update Jetpack script to be async and deferred

jameswburke:
## Summary
As titled.

## Notes for reviewers
None.

## Changelog entries
* **Changed**: Updated Jetpack stats integration to be async and deferred.

## Ticket(s)
None.

### DIFF
--- a/inc/integrations/class-jetpack.php
+++ b/inc/integrations/class-jetpack.php
@@ -54,7 +54,9 @@ class Jetpack {
 					'script',
 					[
 						'config' => [
-							'src' => 'https://stats.wp.com/e-' . gmdate( 'YW' ) . '.js',
+							'async' => 'async',
+							'defer' => 'defer',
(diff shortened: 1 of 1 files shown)


inc/integrations/class-jetpack.php
```diff
@@ -54,7 +54,9 @@ public function inject_jetpack_stats_script_head_children( array $children, arra
 					'script',
 					[
 						'config' => [
-							'src' => 'https://stats.wp.com/e-' . gmdate( 'YW' ) . '.js',
+							'async' => 'async',
```
joemcgill, line 57: Makes me wonder if those should be the default for our integrations somehow?

jameswburke: In WP Components we had some helper methods like that. I'm not opposed to reintroducing a decent `<head>`/`<Helmet>` API.

PR #302, opened 2020-12-08
ensure nonexistent URLs return a 404 instead of the homepage

xmattus:


joemcgill (approved):
Clever! Seems to be working 👍

PR #303, opened 2020-12-09
Fix ability to override theme in a config callback

amyevans:
Refs: [IRV-909](https://alleyinteractive.atlassian.net/browse/IRV-909)

This fixes a bug where
```php
Components\register_component_from_config(
	__DIR__ . '/component',
	[
		'config_callback' => function( array $config ): array {
			$config['theme'] = 'my_new_theme';
			return $config;
		},
	]
);
```

resulted in
```json
...
"config": {
  "themeName": "default",
},
...
```

Now, it will result in
```json
...
"config": {
  "themeName": "myNewTheme",
},
...
```
as expected

### DIFF
--- a/inc/components/class-component.php
+++ b/inc/components/class-component.php
@@ -199,7 +199,7 @@ class Component implements JsonSerializable {
 			->apply_context()
 			->set_config( $args['config'] )
 			->hydrate_config()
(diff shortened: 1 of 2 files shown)


jameswburke (approved):
:bee:

PR #307, opened 2020-12-24
Update site-menu component to allow explicitly setting the menu name

jameswburke:
## Summary
As titled.

## Notes for reviewers
None.

## Changelog entries
* **Changed**: Use menu ID to get the menu items, instead of the menu `WP_Term` name.
* **Fixed**: If `$config['menu_name']` is explicitly set, don't override it.

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/IRV-882

### DIFF
--- a/inc/components/components/site-menu/component.php
+++ b/inc/components/components/site-menu/component.php
@@ -52,6 +52,11 @@ register_component_from_config(
 
 			$menu_items = wp_get_nav_menu_items( $config['menu_id'] );
 
+			// Validate $menu_items is an array of items.
+			if ( ! is_array( $menu_items ) ) {
+				return $children;
+			}
+
 			/*
 			 * Sort the menu items into two groups: top level items, and a list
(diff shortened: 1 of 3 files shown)


inc/components/components/site-menu/component.php
```diff
@@ -47,7 +50,7 @@
 				return $children;
 			}
 
-			$menu_items = wp_get_nav_menu_items( $config['menu_name'] );
+			$menu_items = wp_get_nav_menu_items( $config['menu_id'] );
```
efuller, line 53: Suggestion: Lacking context here, but `wp_get_nav_menu_items` can also return `false`. Perhaps a change to something like `$menu_items = wp_get_nav_menu_items( $config['menu_id'] ) ?? [];` would guarantee `$menu_items` would always be an array.

efuller (approved):
🌮  - Just one minor suggestion/consideration.

PR #310, opened 2020-12-28
Cache full site theme value to improve performance

amyevans:
Generating the full site theme via `get_site_theme_from_json_files()` (which finds files in a directory, loads the contents from JSON, and compiles to an array) each time `get_site_theme()` was called was getting pretty expensive (given that it might get called 1000+ times per page load). Caching the site theme value in a transient for even a very short duration makes local dev in particular much snappier.

### DIFF
--- a/inc/templates/site-theme.php
+++ b/inc/templates/site-theme.php
@@ -54,7 +54,7 @@ function get_site_theme( string $selector = '', $default = null ) {
 
 	// Get cached site theme value, if available.
 	$cache_key = 'wp_irving_site_theme';
-	$theme     = get_transient( $cache_key );
+	$theme     = wp_cache_get( $cache_key );
 	if ( false === $theme ) {
 		/**
 		 * Filter to modify the site theme.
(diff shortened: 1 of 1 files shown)


dlh01 (approved):
🌳 

Assuming Irving uses VIP Go's 60-second Varnish cache TTL for REST API requests, does the TTL here need to be lower than the TTL in Varnish? Perhaps the TTL could be environment-specific?

I'm guessing that any request to the origin would cause this transient to be loaded, in which case, there probably will be some uncached request that uses it within that 30 seconds. Still, noting from experience that on Go, a 30-second TTL might pepper Memcached and draw some attention if cache performance becomes a problem.

joemcgill:
I'm noticing this issue this evening as well. Realistically, I'm not sure we even need to store these values as a transient in the database, since there are so many other cache opportunities after this value is calculated, a `wp_cache_set()` may be fine here just to ensure that the value is only calculated once per PHP run (i.e., when not hitting the varnish cache).

I know that you're doing some trickery with dynamically applying/removing filters in order to merge site theme data from multiple sources, will this end up messing up anything you're doing on that front?

PR #311, opened 2021-01-15
Update Pico staging URL in our integration

joemcgill:
The new domain since Pico v1.0.0 is now `https://gadget.staging.pico.tools`.


amyevans (approved):
✨

PR #312, opened 2021-01-15
Decode HTML entities for title in social sharing component

amyevans:


tests/components/test-components.php
```diff
@@ -883,7 +883,7 @@ public function test_component_post_social_sharing() {
 					],
 					'platformData' => [],
 					'postId'       => $this->get_post_id(),
-					'title'        => get_the_title( $this->get_post_id() ),
+					'title'        => 'It’s like this & that.',
```
joemcgill, line 886: Do we need to update the `get_the_title()` helper to make encoding optional?

joemcgill (approved):
Are there cases where the title property is used where it _should_ be encoded? If not, then I think this is fine. 👍🏻

tests/components/test-components.php
amyevans: Hmm, there isn't an existing helper, we're just using the core function. I can add a helper though.

amyevans:
@joemcgill I added a few more commits - could you take another look?

joemcgill (approved):
👍🏻 changes look good to me

PR #316, opened 2021-02-09
Move irving/head to providers

ostowe:
[**IRV-977**](https://alleyinteractive.atlassian.net/browse/IRV-977)

NOTE: this will break without https://github.com/alleyinteractive/irving/pull/462 in place as well.


kaitlinbolling (approved):
👍

PR #318, opened 2021-02-15
Revert "Update Pico staging URL in our integration"

ostowe:
Reverts alleyinteractive/wp-irving#311


joemcgill (approved):
👍🏻 This is the only blocking change in the WP Irving plugin. Another approach that I think you could take is to programmatically inject the `staging` part of the URL into the endpoint so that it doesn't matter what subdomain is being loaded by the Pico plugin.

ostowe:
ohhh good call, I'll take a look at that

PR #321, opened 2021-02-19
Revert "Remove check for page context in block library styles"

ostowe:
Reverts alleyinteractive/wp-irving#319


jameswburke (approved):
:bee:

PR #324, opened 2021-03-31
Integrate Application Passswords

ostowe:
[**IRV-800**](https://alleyinteractive.atlassian.net/browse/IRV-800)


inc/integrations/class-auth.php
```diff
@@ -0,0 +1,258 @@
...
+		$app_pass_request->set_query_params(
+			[
+				'name'   => $app_name,
+				'app_id' => wp_generate_uuid4(),
+			]
+		);
+
+		// Let's get the token.
+		$result = rest_do_request( $app_pass_request );
+		$status = $result->status ?? 0;
```
renatonascalves, line 250: This is also available.

```suggestion
		$status = $result->get_status() ?? 0;
```

renatonascalves (approved):
🗡️

PR #327, opened 2021-04-13
Bypass auth errors and return unauthenticated data if app pass is incorrect

ostowe:
[**LEDE-1431**](https://alleyinteractive.atlassian.net/browse/LEDE-1431)


jameswburke (approved):
:bee:

PR #329, opened 2021-04-27
Display a message to users with an invalid auth token

jameswburke:
## Summary
Displays a message to users with expired auth sessions.

## Notes for reviewers
None.

## Changelog entries
* **Added**: Message for expired auth sessions.
* **Added**: Ability to refresh tokens via a shortcut link or cookie.

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/LEDE-1453


amyevans (approved):
✨

PR #332, opened 2021-05-12
Prepare changelog for 0.9.0

jameswburke:
* Fix: Update the site theme to work for child themes
* Fix: Remove check for page context in block library styles
* Fix: Update New Relic integration to better typecast
* Add: Integrate Application Passswords
* Add: Initial support for FB Instant Articles
* Fix: Bypass auth errors and return unauthenticated data if app pass is incorrect
* Add: Display a message to users with an invalid auth token
* Fix: Update WP Irving integration to only delete Irving app passwords
* Remove: Feat/deprecate theme options

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,22 @@ While WP Irving contains a few basic plugin integrations (and we hope to support
 
 ## Changelog ##
 
+### 0.10.0 ###
+
+[in development]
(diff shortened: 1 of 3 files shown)


amyevans (approved):
✨

PR #334, opened 2021-05-26
prevent favicon requests from affecting cookies

jakewrfoster:
**_This PR may not address the root cause of the issue._**

I am seeing a situation while using Application Passwords where favicon requests are unsetting domain authorization cookies. The code in this PR prevents favicon requests from unsetting cookie values, but the cause of the issue might be further up in the request lifecycle.

WP-Irving:  v0.9.0
@irvingjs/core: "6.2.0
@irvingjs/integrations: ^6.2.0

### DIFF
--- a/inc/integrations/class-application-passwords-auth.php
+++ b/inc/integrations/class-application-passwords-auth.php
@@ -89,6 +89,10 @@ class Application_Passwords_Auth {
 	 * Handle the cookie logic upon init.
 	 */
 	public function handle_cookie() {
+		// Prevent Favicon requests from affecting cookies.
+		if ( isset( $_SERVER['REQUEST_URI'] ) && '/favicon.ico' === $_SERVER['REQUEST_URI'] ) {
+			return;
+		}
 
 		/**
 		 * Determine the cross domain cookie domain.


jakewrfoster:
https://user-images.githubusercontent.com/1018205/119600960-2a852b00-bdb6-11eb-9f97-d94259f684ab.mp4

jakewrfoster:
Awaiting additional reviews before merging!

jameswburke:
I'd like to replicate this issue in a vanilla WordPress and Irving install before we move forward on this.

jameswburke:
@jakewrfoster if you're looking to fix this on a project, I'd recommend unhooking the actions in `setup()` and wrapping them with whatever conditional(s) you need.

jakewrfoster:
@jameswburke 
image

The callback function is hashed, so I don't think that I can unhook the action, would you be satisfied with adding a filter in the `handle_cookies` callback that could allow applications to bail early? Some thing like:

```php
/**
 * Handle the cookie logic upon init.
 */
 public function handle_cookie() {
    if ( ! apply_filters( 'wp_irving_app_pw_should_handle_cookies', true ) ) {
        return;
    }
```

jakewrfoster:
ah, disregard, the above. I can remove the hook by instancing the class and removing it that way:

```php
add_action( 'plugins_loaded', function () {
	if ( isset( $_SERVER['REQUEST_URI'] ) && '/favicon.ico' === $_SERVER['REQUEST_URI'] ) {
		$app_pw_class = \WP_Irving\Integrations\Application_Passwords_Auth::instance();
		remove_action( 'init', [
			$app_pw_class,
			'handle_cookie'
		] );
	};
} );
```

jakewrfoster:
closing this for now, thanks!

PR #335, opened 2021-06-17
Refactor Post List to deduplicate posts more performantly

jameswburke:
## Summary
As titled.

## Notes for reviewers
Docs on why and how are here, https://docs.wpvip.com/technical-references/code-quality-and-best-practices/using-post__not_in/

## Changelog entries
* **Changed**: Refactor `irving/post-list` to not use `post__not_in` for deduplication purposes.

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/IRV-1003

### DIFF
--- a/inc/components/components/post-list/component.php
+++ b/inc/components/components/post-list/component.php
@@ -18,22 +18,17 @@ register_component_from_config(
 	__DIR__ . '/component',
 	[
 		'config_callback'   => function ( array $config ): array {
+
+			// Grab the global in case we need it.
 			global $wp_query;
-			$query = $wp_query;
 
-			$query_args = $config['query_args'] ?? [];
-
-			if ( ! empty( $query_args ) ) {
-
-				if ( wp_validate_boolean( $query_args['exclude'] ?? false ) ) {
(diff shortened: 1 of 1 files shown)


amyevans (approved):
✨ 

Curious to know what performance improvements you saw after implementing this. I'm a bit worried that this could still end up making pretty heavy requests (e.g. for like 100 posts), and not a consistent number as to result in cache hits (per VIP's documentation: `Make it a fixed number, not variable, to reduce the number of cache variants.`)... but I think it's still an improvement 😄

jameswburke:
> Curious to know what performance improvements you saw after implementing this. I'm a bit worried that this could still end up making pretty heavy requests (e.g. for like 100 posts), and not a consistent number as to result in cache hits (per VIP's documentation: `Make it a fixed number, not variable, to reduce the number of cache variants.`)... but I think it's still an improvement 😄

In my original scoping, I was seeing around 50% increase for something like a Lede homepage.

For max performance, I agree, we're still not there, but it's such a tricky problem to solve. I have some other thoughts we can chat through.

PR #336, opened 2021-06-28
Improve Coral integration to account for permalink modifications

amyevans:
Currently, if a post's permalink is changed, a new Coral commenting thread is created because Coral is using the post's permalink as the only unique identifier. This PR:

* Explicitly passes the post ID in the Coral embed as the unique story identifier
* Updates story metadata in Coral when the permalink is changed

These changes ensure the existing commenting thread is used on a post with a modified permalink.

### DIFF
--- a/inc/integrations/class-coral.php
+++ b/inc/integrations/class-coral.php
@@ -665,7 +665,7 @@ class Coral {
 		}
 
 		// Something else went wrong; email the result to the site admin.
-		wp_mail(
+		wp_mail( /* phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_mail_wp_mail */
 			get_option( 'admin_email' ),
 			__( 'Unsuccessful Coral Story URL Update', 'wp-irving' ),
 			sprintf(
--- a/inc/integrations/class-coral.php
+++ b/inc/integrations/class-coral.php
@@ -70,6 +70,9 @@ class Coral {
 		}
 
 		add_action( "save_post_{$this->post_type}", [ $this, 'delete_cached_values' ], 10, 2 );
+
+		// Add hook for updated permalinks so that changes can be passed along to Coral.
+		add_action( 'added_post_meta', [ $this, 'check_for_updated_permalink' ], 10, 3 );
 	}
 
 	/**
@@ -89,6 +92,22 @@ class Coral {
 			'wp_irving_coral_sso_secret',
 			esc_html__( 'Coral SSO Secret', 'wp-irving' ),
 			[ $this, 'render_coral_sso_secret_input' ],
+			'wp_irving_integrations',
+			'irving_integrations_settings'
+		);
+
+		add_settings_field(
+			'wp_irving_coral_admin_email',
+			esc_html__( 'Coral Admin Email', 'wp-irving' ),
+			[ $this, 'render_coral_admin_email_input' ],
+			'wp_irving_integrations',
+			'irving_integrations_settings'
+		);
+
+		add_settings_field(
+			'wp_irving_coral_admin_password',
+			esc_html__( 'Coral Admin Password', 'wp-irving' ),
+			[ $this, 'render_coral_admin_password_input' ],
 			'wp_irving_integrations',
 			'irving_integrations_settings'
 		);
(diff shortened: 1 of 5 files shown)


inc/integrations/class-coral.php
```diff
@@ -584,4 +627,162 @@ public function delete_cached_values( $post_id, $post ) {
...
+		}
+
+		// Validate Coral is in use.
+		$coral_url = untrailingslashit( Integrations\get_option_value( 'coral', 'url' ) );
+		if ( empty( $coral_url ) ) {
+			return;
+		}
+
+		// Attempt to update the permalink.
+		$response = $this->update_permalink_in_coral( $object_id );
```
jameswburke, line 653: The only remaining edge case that comes to mind is handling posts that aren't using the post id as the unique coral ID (basically, any post that exists before these PRs merge to production).

So what happens if the `id: {$post_id},` is not found in the graphql request? Does everything fail gracefully?

inc/integrations/class-pico.php
```diff
@@ -144,14 +144,14 @@ public function render_pico_auth_key_input() {
...
-	 * Render an input for the Pico Whitelisted SSO Tiers.
+	 * Render a textarea for the Pico allow listed SSO Tiers.
 	 */
 	public function render_pico_tiers_input() {
 		// Check to see if there are existing tiers in the option.
 		$tiers = $this->options[ $this->option_key ]['tiers'] ?? '';
 
 		?>
-			<input id="pico_tiers" type="text" name="irving_integrations[<?php echo esc_attr( 'pico_tiers' ); ?>]" value="<?php echo esc_attr( $tiers ); ?>" />
+			<textarea id="pico_tiers" rows="5" cols="50" type="text" name="irving_integrations[<?php echo esc_attr( 'pico_tiers' ); ?>]"><?php echo esc_attr( $tiers ); ?></textarea>
```
jameswburke, line 154: Thank you. One other quick thing, could we trim whitespace + any trailing commas on this input during save or fetch? I want to handle all these situations,

- `'one,two,three' => [ 'one', 'two', 'three' ]` 👍 
- `'one, two, three' => [ 'one ', 'two ', 'three ' ]` 👎 
- `'one, two, three, ' => [ 'one', 'two', 'three', '']` 👎 
- `'one, two, three,' => [ 'one', 'two', 'three', '']` 👎 

I think an explode( ',', $values) and array_values() should just about handle it?

Could also be a new ticket if we prefer.

inc/integrations/class-coral.php
```diff
@@ -584,4 +627,162 @@ public function delete_cached_values( $post_id, $post ) {
...
 	}
+
+	/**
+	 * Hook into added post meta and identify additions of `_wp_old_slug`.
+	 * Pass along the updated permalink to Coral so that the post continues to
+	 * be properly identified.
+	 *
+	 * @param int    $meta_id     ID of updated metadata entry.
+	 * @param int    $object_id   ID of the object metadata is for.
+	 * @param string $meta_key    Metadata key.
```
jameswburke, line 638: Spacing adjustment,
```suggestion
	 * @param int    $meta_id   ID of updated metadata entry.
	 * @param int    $object_id ID of the object metadata is for.
	 * @param string $meta_key  Metadata key.
```

jameswburke (approved):
:bee:

inc/integrations/class-coral.php
amyevans: Yeah, I tested under this scenario and it fails gracefully. The graphql request returns
```
{
  "data": {
    "updateStory": {
      "story": null
    }
  }
}
```
if the ID doesn't exist.

But what really happens here is accessing the post at the new permalink leads to the Lazy Story creation of another entry in Coral (so new comment thread). This is the same behavior as before, so I felt like it was enough of an edge case to not bother addressing. Any newly created posts after this merge will behave as we want, any preexisting posts will have preexisting behavior.

inc/integrations/class-pico.php
amyevans: Ah yes, thanks for the reminder, this has bugged me in the past too. Fixing

inc/integrations/class-coral.php
jameswburke: Right, that sounds good to me.

PR #338, opened 2021-07-01
Only send Story ID for published posts

amyevans:
Sending the story ID for a Coral embed on post previews causes issues


cahdeemer (approved):
🌻

PR #340, opened 2021-07-16
Allow filter to handle generic objects instead of WP_Term objects

jameswburke:
## Summary
As titled.

## Notes for reviewers
None.

## Changelog entries
* **Fixed**: Address an error in the type hinting for the term row actions.

## Ticket(s)
None.


cahdeemer (approved):
🌻

PR #342, opened 2021-07-28
Add support for additional Coral config and account for API fetch limit of 25

amyevans:
This PR:
* Adds `article_URL` and `no_text` config to achieve parity with original Coral component (see companion PR: https://github.com/alleyinteractive/irving/pull/590)
* Fixes the logic for the cron API calls. Previously I had assumed requesting more stories than were available was causing the Internal Error response, but after additional testing I realized the limit is 25 (this is undocumented), regardless of how many stories are available. So now we'll request all stories on the first run, then the 25 most recently active every 15 minutes, and if more than 25 were active in those 15 minutes, perform another full story fetch. Not the most ideal, but also probably unlikely to happen that often.

### DIFF
--- a/inc/components/components/coral-comment-count-static/component.json
+++ b/inc/components/components/coral-comment-count-static/component.json
@@ -2,6 +2,10 @@
   "name": "irving/coral-comment-count-static",
   "description": "Coral comment count that is fetched from post meta.",
   "config": {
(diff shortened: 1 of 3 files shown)


jameswburke (approved):
:bee:

PR #346, opened 2021-08-02
Add action hook for logging 401 Pico API errors

amyevans:
This PR adds a mechanism for hooking in and logging/alerting re: 401 API errors


jameswburke (approved):
:bee:

Will 401 only occur if the key is wrong? What if the user's creds are wrong?

amyevans:
If the user creds are wrong, it's also a 401 (with a different `message`), so valid point. I think I'll handle filtering out what we don't want to log within the callback though.

PR #350, opened 2021-08-24
Ensure Coral user post type actually gets registered

amyevans:
The integrations setup is running on `init`, so attempting to hook on `init` here was being ignored


jameswburke (approved):
:bee:

PR #354, opened 2021-09-22
Don't camel case CSS custom properties

ostowe:


amyevans (approved):
✨  Looks good - though possible to add a test case for the scenario?

ostowe:
hmm yeah good point, I can do that

PR #355, opened 2021-09-24
Add coral comment count threshold setting

jameswburke:
## Summary
Introduces a new Coral integration field for comment count threshold.

## Notes for reviewers
By itself, this value doesn't do anything - it's up to individual projects to utilize it.

## Changelog entries
* **Added**: Number field in the Irving integrations for setting a Coral comment count icon threshold value.

## Ticket(s)
https://alleyinteractive.atlassian.net/browse/LEDE-1915

### DIFF
--- a/inc/integrations/class-coral.php
+++ b/inc/integrations/class-coral.php
@@ -138,6 +138,14 @@ class Coral {
 		);
 
 		add_settings_field(
+			'wp_irving_coral_comment_count_display_threshold',
+			esc_html__( 'Minimum number of comments required to display the comment count', 'wp-irving' ),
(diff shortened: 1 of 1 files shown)


inc/integrations/class-coral.php
```diff
@@ -220,6 +228,18 @@ public function render_coral_use_cron_input() {
 		<?php
 	}
 
+	/**
+	 * Render a number input which determines when comment count icons appear.
+	 */
+	public function render_coral_comment_count_display_threshold_input() {
+		// Check to see if there is an existing SSO secret in the option.
```
amyevans, line 235:
```suggestion
		// Check to see if there is an existing value.
```

amyevans (approved):
✨

PR #359, opened 2021-12-02
Only render an admin bar component for logged in users

amyevans:
In [this commit](https://github.com/alleyinteractive/wp-irving/commit/a5df5c6618fe8757aae21b8d135456b76f34eaea#diff-66dcd42df3bee74d79853351a5473e8b19b9a4c9cb8dacee87f9726b50ebef79), logic was removed that prevented the admin bar component from being included in the page data for logged out users. This restores that logic (but still leaves out the "Looks like your session has expired." language).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,8 @@
     }
   ],
   "require-dev": {
-    "alleyinteractive/alley-coding-standards": "dev-main"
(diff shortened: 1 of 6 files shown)


ostowe (approved):
💯

amyevans:
I've created a separate issue to track and fix the PHPUnit failure

PR #360, opened 2021-12-09
Modify preview class to remove JWT references

amyevans:
Application Passwords is now the default (and is part of WordPress core)


ostowe (approved):
💯

PR #362, opened 2022-10-05
use method_exists instead of is_callable to check for SRM

ostowe:
In PHP8 `is_callable` doesn't behave the same way we'd expect for our integrations functionality. This PR switches to using `method_exists` instead.


vaguity (approved):
💯

PR #364, opened 2024-01-17
add missing variable declaration

matthewmcvickar:
The WP Irving plugin is showing an error:

```
PHP Deprecated:  Creation of dynamic property WP_Irving\JWT_Auth::$cookie_domain is deprecated in /wp/wp-content/client-mu-plugins/wp-irving/inc/integrations/class-jwt-auth.php on line 88
```

This is happening because the `WP_Irving\JWT_Auth` class does not declare the `$cookie_domain` variable before using it.

Not declaring this variable before it's used causes a 'dynamic variable' to be created, which is [deprecated in PHP 8.2](https://php.watch/versions/8.2/dynamic-properties-deprecated).

This PR fixes that by declaring the variable.


matthewmcvickar:
Sorry; this was meant to be opened in a different fork of this plugin, not this one. Please ignore.

